PR #24, opened 2022-06-14
Allow to repeat slave detection step in case a slave is not ready.

### LINKED ISSUE #23: detectSlave api
It would be nice that in the bus class, detectSlave() return directly the number of detected slaves instead of having to call for the getter afterward.

Rdk-T:
close #23 

Also add a new Exception type to handle Datagram Error and expose the datagram state to the client.


Rdk-T:
How do you generate code coverage of the project ?

Siviuze:
By enabling CODE_COVERAGE option directly in CMake

PR #30, opened 2022-09-08
Moving prints to Print.h

Eyusd:
This MR tidies things a bit, moving printing functions into Prints.h
Tests for prints are added, though not included in the code coverage


CMakeLists.txt
```diff
@@ -75,6 +77,7 @@ if (GTest_FOUND)
     include(CodeCoverage)
     append_coverage_compiler_flags()
     set(GCOVR_ADDITIONAL_ARGS "--exclude-unreachable-branches")
+    set(COVERAGE_EXCLUDES "src/Prints.cc")
```
Rdk-T, line 80: l84 there is already the Exclude list, you can move this file to this line.

PR #31, opened 2022-09-09
Prints moved to Prints.h

Eyusd:
This MR tidies things up a bit.
Printing methods are moved to Prints.h, and removed from code coverage reports.

### DIFF
--- a/include/kickcat/Prints.h
+++ b/include/kickcat/Prints.h
@@ -10,14 +10,14 @@
 namespace kickcat
 {
     // Slaves utils
-    void printInfo(Slave slave);
-    void printPDOs(Slave slave);
-    void printErrorCounters(Slave slave);
-    void printDLStatus(Slave slave);
-    void printGeneralEntry(Slave slave);
+    void printInfo(const Slave& slave);
+    void printPDOs(const Slave& slave);
+    void printErrorCounters(const Slave& slave);
+    void printDLStatus(const Slave& slave);
+    void printGeneralEntry(const Slave& slave);
 
     // Topology utils
-    void printTopology(std::unordered_map<uint16_t, uint16_t> map);
+    void printTopology(const std::unordered_map<uint16_t, uint16_t>& topology_mapping);
 }
 
 #endif
--- a/include/kickcat/Prints.h
+++ b/include/kickcat/Prints.h
@@ -1,0 +1,23 @@
+#ifndef KICKCAT_PRINTS_H
+#define KICKCAT_PRINTS_H
+
+#include <unordered_map>
+#include <cstdint>
+
+#include "Slave.h"
+
+
+namespace kickcat
+{
+    // Slaves utils
+    void printInfo(Slave slave);
+    void printPDOs(Slave slave);
+    void printErrorCounters(Slave slave);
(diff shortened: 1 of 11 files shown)


include/kickcat/Prints.h
```diff
@@ -0,0 +1,23 @@
...
+#include <unordered_map>
+#include <cstdint>
+
+#include "Slave.h"
+
+
+namespace kickcat
+{
+    // Slaves utils
+    void printInfo(Slave slave);
```
Siviuze, line 13: const& everywhere (avoid unnecessary copy)

include/kickcat/Prints.h
```diff
@@ -0,0 +1,23 @@
...
+    void printPDOs(Slave slave);
+    void printErrorCounters(Slave slave);
+    void printDLStatus(Slave slave);
+    void printGeneralEntry(Slave slave);
+
+    // Topology utils
+    void printTopology(std::unordered_map<uint16_t, uint16_t> map);
```
Siviuze, line 20: parameter name is meaningless (the type is already a map).

src/Prints.cc
```diff
@@ -0,0 +1,92 @@
+#include "Prints.h"
+
+namespace kickcat
+{
+    void printInfo(const Slave& slave)
```
Rdk-T, line 5: To keep the consistency with the convention of the rest of the project prefer using `Slave const&` see what is done in Frame.h or Bus.h.

The code will be working the same, it's just a convention since the rule is "const applies to the thing left of it. If there is nothing on the left then it applies to the thing right of it."

src/Prints.cc
```diff
@@ -0,0 +1,92 @@
...
+            printf( "port_2: %04x \n",                  general_entry->port_2);
+            printf( "port_3: %04x \n",                  general_entry->port_3);
+        }
+    }
+
+
+    void printTopology(std::unordered_map<uint16_t, uint16_t> const& topology_mapping)
+    {  
+        printf( "\n -*-*-*-*- Topology -*-*-*-*-\n" );
+        for (auto& it : topology_mapping)
```
Siviuze, line 80: auto const&

src/Prints.cc
```diff
@@ -0,0 +1,92 @@
...
+        printf("  Communications : %d \n", slave.dl_status.COM_port2);
+        printf("  Loop Function :  %d \n", slave.dl_status.LOOP_port2);
+
+        printf("Port 3: \n");
+        printf("  Physical Link :  %d \n", slave.dl_status.PL_port3);
+        printf("  Communications : %d \n", slave.dl_status.COM_port3);
+        printf("  Loop Function :  %d \n", slave.dl_status.LOOP_port3);
+    }
+
+    void printGeneralEntry(Slave const& slave) {
```
Siviuze, line 44: line feed

src/Prints.cc
```diff
@@ -0,0 +1,92 @@
...
+        eeprom::GeneralEntry const* general_entry = slave.sii.general;
+        
+        if (general_entry == NULL)
```
Siviuze, line 47: nullptr in C++, not NULL

Siviuze (approved):

PR #37, opened 2022-09-20
Add debughelpers

Eyusd:
This MR proposes ready-to-use functions to read and write registers on an EtherCat device, used in an example.
Refer to device documentation to know whether a specific memory slot is available for R/W operations.


tools/debugExample.cc
```diff
@@ -0,0 +1,54 @@
...
+        return 1;
+    }
+
+    auto socket = std::make_shared<Socket>();
+    Link link(socket);
+
+    try
+    {
+        socket->open(argv[1], 2ms);
+        socket->setTimeout(1ms);
```
Rdk-T, line 32: You can directly set 1ms at the open call (the 2ms parameter sets the socket's timeout as well)

unit/debug-t.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "Mocks.h"
+
+using ::testing::Return;
+using ::testing::_;
+using ::testing::Invoke;
+using ::testing::InSequence;
+
+using namespace kickcat;
+
+struct SDOAnswer
```
Rdk-T, line 15: This struct is unused

unit/debug-t.cc
```diff
@@ -0,0 +1,132 @@
...
+{
+    mailbox::Header header;
+    mailbox::ServiceData sdo;
+    uint8_t payload[4];
+} __attribute__((__packed__));
+
+class FullLink : public testing::Test
```
Rdk-T, line 22: What do you mean by "FullLink" ?

unit/debug-t.cc
```diff
@@ -0,0 +1,132 @@
...
+{
+public:
+
+
+
+    template<typename T>
+    void handleReply(std::vector<T> answers, uint16_t replied_wkc = 1)
```
Rdk-T, line 29: The `FullLink` class seems to be an extract of the `BusTest` class. To avoid code duplication it may be better to move the tests `send_get_register` and `send_write_register` into bus-t

What do you think @Siviuze ?

tools/CMakeLists.txt
```diff
@@ -1,10 +1,21 @@
-set(EXAMPLE_SRCS scanTopology.cc)
+set(EXAMPLE_TOPO scanTopology.cc)
```
Siviuze, line 1: strange name for sources

tools/debugExample.cc
```diff
@@ -0,0 +1,54 @@
+#include "kickcat/Link.h"
```
Siviuze, line 1: This is the tool folder, not the example section

unit/debug-t.cc
Siviuze: Yes, that or create another file that can be use by both bus-t and debug-t. Note: a better approach seems to be to mock the Link class to avoid managing the whole communication. It may be worth to wait for your Link refactoring.

PR #39, opened 2022-09-29
RFC: Link redundancy

### LINKED ISSUE #38: Add redundancy compatibility
Rework the code base to allow the use of redundancy.

Rdk-T:
close #38 

Todo:

- [ ] rework fusion frame
- [ ] update archi doc


include/kickcat/Frame.h
```diff
@@ -49,21 +49,25 @@ namespace kickcat
 
         /// \return true if datagram can be extracted, false otherwise
         bool isDatagramAvailable() const { return is_datagram_available_; }
+        void setIsDatagramAvailable(bool value) { is_datagram_available_ = value; }
```
Siviuze, line 52: You may force this accessor to always set the flag to true since there is no other needs.

src/Prints.cc
```diff
@@ -90,4 +90,16 @@ namespace kickcat
             }
         }
     }
+
+
+    void printDatagramHeader(DatagramHeader const& header)
```
Siviuze, line 95: in a dedicated PR?

unit/Mocks.h
```diff
@@ -12,5 +24,71 @@ namespace kickcat
...
+        void readError()
+        {
+            EXPECT_CALL(*this, read(::testing::_, ::testing::_))
+            .WillOnce(::testing::Invoke([](uint8_t* , int32_t)
+            {
+                return -1;
+            }));
+        }
+
+        // TODO QUEUE
```
Siviuze, line 91: in a future PR?

Siviuze:
No more comments on 'style'. On content however, here after the point to rework (as already said in our previous discussion):
- remove AbstractLink and SimpleLink and always use the RedundantLink even without redundancy (it will reduce maintenance load and it should not add cost for users that do not need redundancy with your implementation)
- rework mocks helpers to improve unit tests

PR #45, opened 2022-11-04
move prints to prints file

Rdk-T:
related https://github.com/Siviuze/KickCAT/issues/40

Went for a to_string approach so that the struct descriptions can be used by remote requests.


include/kickcat/protocol.h
```diff
@@ -167,6 +168,7 @@ namespace kickcat
             LOOP_port3 : 1,
             COM_port3 : 1;
     }__attribute__((__packed__));
+    std::string toString(DLStatus counters);
```
Siviuze, line 171: const&

include/kickcat/protocol.h
```diff
@@ -332,6 +334,8 @@ namespace kickcat
         } __attribute__((__packed__));
     }
```
Siviuze, line 336: line feed

Siviuze (approved):

PR #53, opened 2022-11-29
share the timeout between the two sockets

### LINKED ISSUE #52: Have a common timeout for both interfaces (redundancy related)

Rdk-T:
close #52

### DIFF
--- a/src/Link.cc
+++ b/src/Link.cc
@@ -120,6 +120,8 @@ namespace kickcat
             DEBUG_PRINT("Fail to write on redundancy interface \n");
         }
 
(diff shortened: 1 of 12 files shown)


src/Link.cc
```diff
@@ -141,6 +141,8 @@ namespace kickcat
 
     void Link::writeThenRead(Frame& frame)
     {
+        socket_nominal_->setTimeout(duration_cast<microseconds>(timeout_));
```
Siviuze, line 144: Seems easier to migrate everything on nanoseconds to avoid casting everywhere

PR #69, opened 2023-01-16
Allow client to override timeout callback of wait for state.

### LINKED ISSUE #68: Allow client to override waitForState timeout callback
use case: when a slave is slower to switch from safe op to op than the PDO watchdog set, it leads the slave to trigger its internal PDO watchdog.

To avoid this, as a client, we need to send multiple PDO while waiting for the switch to OP to prevent the watchdog to be triggered.

Rdk-T:
close #68


src/Bus.cc
```diff
@@ -164,7 +166,7 @@ namespace kickcat
 
             if (elapsed_time(now) > timeout)
             {
-                THROW_ERROR("Timeout");
+                THROW_ERROR("Timeout \n");
```
Siviuze, line 169: Why adding a linefeed?

Siviuze (approved):

src/Bus.cc
Rdk-T: my bad

PR #76, opened 2023-02-06
fix possible warnings & snake_case

eyusd:
Fix possible warnings (-Wconversion -Wunused-parameter) in emitter.cc, and convert server.cc to snake_case

### DIFF
--- a/examples/test/emitter.cc
+++ b/examples/test/emitter.cc
@@ -16,10 +16,8 @@
 
 using namespace kickcat;
 
-int main(int argc, char* argv[])
+int main(int, char**)
 {
-    (void) argc;
-    (void) argv;
     printf("Start\n");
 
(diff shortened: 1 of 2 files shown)


examples/test/emitter.cc
```diff
@@ -62,17 +60,15 @@ int main(int argc, char* argv[])
         msg->setAddress(1001);
 
         std::memcpy(frame + sizeof(EthercatHeader), msg->data(), msg->size());
-        header->len = msg->size();
+        header->len = static_cast<uint16_t>(msg->size() & 2047); // Ensures that size() won't use more than 11 bits
```
Siviuze, line 63: 2047 -> prefer hexadecimal representation. Not sure if the static_cast is needed with the mask.

Siviuze:
No activity on it: closed.

PR #111, opened 2023-08-03
update dl status to deal with pdi watchdog status

### LINKED ISSUE #110: do not process datagram when calling `sendGetDLStatus`
https://github.com/Siviuze/KickCAT/blob/84fded65c8f95908f7be97e97ee6e087cfaaf3c2/src/Bus.cc#L1005

In the case where we want to fetch the DL status of multiple slaves, with the current code we will have to for loop through all the slaves sending datagrams one at a time vs just having one process datagram with the request dl status for all n slaves

### LINKED ISSUE #112: Update DL status to incorporate PDI watchdog
PDI watchdog allows us to determine if the slave application has crashed/frozen as the esc lives on to communicate with the master even thought the slave app is probably dead
Screenshot from 2023-08-03 11-04-11

trns1997:
Closes #110
Closes #112

### DIFF
--- a/src/Bus.cc
+++ b/src/Bus.cc
@@ -1002,7 +1002,6 @@ namespace kickcat
         };
 
         link_->addDatagram(Command::FPRD, createAddress(slave.address, reg::ESC_DL_STATUS), nullptr, 2, process, error);
-        link_->processDatagrams();
     }
 
 }
--- a/src/protocol.cc
+++ b/src/protocol.cc
@@ -175,30 +175,30 @@ namespace kickcat
     {
         std::stringstream os;
 
-        os << "PDI Operational:  " << std::to_string(dl_status.PDI_op) << " \n";
-        os << "PDI Watchdog Status : " << std::to_string(dl_status.PDI_watchdog) << " \n";
+        os << "PDI Operational:  "          << std::to_string(dl_status.PDI_op) << " \n";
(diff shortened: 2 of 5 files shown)


src/protocol.cc
```diff
@@ -175,6 +175,11 @@ namespace kickcat
     {
         std::stringstream os;
 
+        os << "PDI Operational:  " << std::to_string(dl_status.PDI_op) << " \n";
```
Siviuze, line 178: not sure that to_string is needed here since this is already a stream that should do the conversion :thinking:

src/Bus.cc
```diff
@@ -1002,7 +1002,6 @@ namespace kickcat
         };
 
         link_->addDatagram(Command::FPRD, createAddress(slave.address, reg::ESC_DL_STATUS), nullptr, 2, process, error);
-        link_->processDatagrams();
```
Siviuze, line 1005: unit tests are broken

trns1997: yes fixing

src/protocol.cc
trns1997: is that true for all the prints here?

Siviuze: probably... to be tested though.

trns1997: Ok sooo we can remove the `to_string` to all variables that are not a `uint8_t`, but in the case a variable is a `uint8_t` the stringstream parses it as a a char and it breaks the print :).

Siviuze: OK

PR #123, opened 2023-09-29
Nuttx slave stack for arduino due

Rdk-T:
close https://github.com/leducp/KickCAT/issues/105


slave_stack/examples/nuttx_arduino_due/main.cc
```diff
@@ -0,0 +1,75 @@
...
+//        reportError(esc.read(0x1000, &test_read, nb_bytes));
+//        for (uint32_t i=0; i < nb_bytes; i++)
+//        {
+//            printf("%x ", test_read[i]);
+//        }
+//        printf(" received\n");
+//    }
+}
+
+extern "C" int main(int argc, char *argv[])
```
trns1997, line 42: I do not think you need extern "C" with @leducp we tested and cpp int main works :)

slave_stack/drivers/nuttx/SPI.cc
```diff
@@ -0,0 +1,34 @@
...
+#include <fcntl.h>
+
+namespace kickcat
+{
+    void SPI::init()
+    {
+        fd_ = file_open(&filep_spi_, "/dev/spi0", O_RDWR);
+        printf("Open spi0 driver file fd: %d \n", fd_);
+
+        priv_spi_ = static_cast<spi_driver_s*>(filep_spi_.f_inode->i_private);
```
leducp, line 12: Try to cast it to  spi_dev_s**

slave_stack/examples/nuttx_arduino_due/main.cc
```diff
@@ -0,0 +1,75 @@
...
+    while(true)
+    {
+        esc_routine(esc);
+        usleep(1000);
+    }
+
+    return 0;
+}
+
+
```
leducp, line 72: trailing

slave_stack/include/kickcat/Error.h
```diff
@@ -6,72 +6,72 @@
 namespace kickcat
 {
     // Errno codes:
-    enum hresult: uint32_t
+    enum class hresult: uint32_t
```
leducp, line 9: Note: we need to settle the error management mechanism.

slave_stack/drivers/nuttx/SPI.cc
Rdk-T: it worked !

@trns1997 so we can get rid of 
```
    struct spi_driver_s
    {
        FAR struct spi_dev_s* spi; /* Contained SPI lower half driver */
    #ifndef CONFIG_DISABLE_PSEUDOFS_OPERATIONS
        mutex_t lock;              /* Mutual exclusion */
        int16_t crefs;             /* Number of open references */
        bool unlinked;             /* True, driver has been unlinked */
    #endif
    };
```

leducp (approved):

PR #137, opened 2024-02-22
Load CANOpen dictionary from ESI file

leducp:


lib/include/kickcat/CoE/OD.h
```diff
@@ -0,0 +1,191 @@
...
+        Entry() = default;
+        ~Entry();
+
+        Entry(Entry const&) = delete;
+        Entry& operator=(Entry const& other) = delete;
+
+        Entry(Entry&& other);
+        Entry& operator=(Entry&& other);
+
+        uint8_t      subindex;
```
Rdk-T, line 166: In table 48 "Get Entry Description Response" ETG1000.6 it specifies also the value info and datatype before bit len.
And no description.

There seems to be changes of order/mismatch with this table.

lib/src/CoE/EsiParser.cc
```diff
@@ -0,0 +1,347 @@
...
+        {
+            if (node_setting->GetText()[0] == '1') { flags |= Access::SETTING; }
+        }
+
+        return flags;
+    }
+
+    std::tuple<DataType, uint16_t> EsiParser::toType(XMLNode* node)
+    {
+        std::unordered_map<std::string, std::tuple<DataType, uint16_t>> BASIC_TYPES
```
Rdk-T, line 192: Why re declare the map at each call ?

lib/include/kickcat/CoE/OD.h
leducp: This entry is NOT the answer of the SDO, but an internal storage: it has to be serialized for such purpose.

Doing this way enable non POD management (so constructor/destructor, std::string or STL container, etc). It seems more important to have a good itnernal class than an already 'mappable' version for SDO Info service.

lib/src/CoE/EsiParser.cc
leducp: Good catch :)

lib/CMakeLists.txt
```diff
@@ -1,9 +1,16 @@
-set(KICKCAT_SOURCES 
+set(KICKCAT_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/src/Error.cc
   ${CMAKE_CURRENT_SOURCE_DIR}/src/protocol.cc
   ${CMAKE_CURRENT_SOURCE_DIR}/src/Time.cc
+
```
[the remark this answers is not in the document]
leducp: voluntary to separate CoE section from the more general section.

Rdk-T (approved):

PR #144, opened 2024-03-13
Add mailbox request

leducp:


lib/include/kickcat/Mailbox.h
```diff
@@ -155,4 +161,69 @@ namespace kickcat
...
+
+        AbstractESC* esc_;
+        SyncManagerConfig mbx_in_;
+        SyncManagerConfig mbx_out_;
+        uint16_t max_msgs_;
+
+        // session handle, from 1 to 7, it is used to detect duplicate frame
+        uint8_t counter_{0};
+
+        ///
```
Rdk-T, line 221: typo

lib/include/kickcat/Mailbox.h
```diff
@@ -155,4 +161,69 @@ namespace kickcat
     };
 }
 
+namespace kickcat::mailbox::response
+{
+    class Mailbox;
```
Rdk-T, line 166: Why the need for the forward declaration ?

lib/include/kickcat/Mailbox.h
```diff
@@ -155,4 +161,69 @@ namespace kickcat
...
+        void replyError(std::vector<uint8_t>&& raw_message, uint16_t code); // wrapper on mailbox replyError
+        uint16_t replyExpectedSize();
+
+        std::vector<uint8_t> data_;
+
+    private:
+        Mailbox* mailbox_;
+    };
+
+    /// \brief
```
Rdk-T, line 195: to complete

Rdk-T:
I find the naming around the mailbox a bit confusing. 

We have a namespace mailbox, with then the namespaces request/response. In these namespaces we have the `struct mailbox` and the `class mailbox`. IMO it would be easier to differentiate them by renaming them explicitly MailboxResponse and MailboxRequest.


There is also the file `CoE/Mailbox`, which only contains the class SDOMessage, I would suggest to rename the file `SDOMessage` instead of mailbox.

lib/slave/src/AbstractESC.cc
```diff
@@ -365,8 +365,9 @@ namespace kickcat
 
     void AbstractESC::set_state_on_error(State state, StatusCode error_code)
     {
+        write(reg::AL_STATUS_CODE, &error_code, sizeof(error_code));
```
Rdk-T, line 368: This looks like a conflict during the rebase

lib/src/CoE/Mailbox.cc
```diff
@@ -0,0 +1,237 @@
...
+
+        reply(std::move(data_));
+    }
+
+    ProcessingResult SDOMessage::process()
+    {
+        auto [object, entry] = findObject(dictionary_, sdo_->index, sdo_->subindex);
+        if (object == nullptr)
+        {
+            printf("Oh no, this object is in another castle! %x.%d\n", sdo_->index, sdo_->subindex);
```
Rdk-T, line 56: Cute, but will be removed in the end ;)

lib/src/Mailbox.cc
```diff
@@ -457,3 +458,159 @@ namespace kickcat
...
+
+    void Mailbox::send()
+    {
+        if (to_send_.empty())
+        {
+            return;
+        }
+
+        auto& msg = to_send_.front();
+        int32_t written_bytes = esc_->write(mbx_out_.start_address, msg.data(), msg.size());
```
Rdk-T, line 571: Is there no check to perform to know if the previous message was properly used by the ESC ?

simulation/ESC.h
```diff
@@ -2,12 +2,13 @@
...
 #include "kickcat/protocol.h"
-#include <vector>
+#include "kickcat/AbstractESC.h"
+
 #include <functional>
 
 namespace kickcat
 {
-    class ESC
+    class ESCEmulator final : public AbstractESC
```
Rdk-T, line 11: Might as well rename the file to be consistent

lib/include/kickcat/Mailbox.h
leducp: Because AbstractMessage need it (line 171) and Mailbox needs also AbstractMessage.

leducp:
> I find the naming around the mailbox a bit confusing.
> 
> We have a namespace mailbox, with then the namespaces request/response. In these namespaces we have the `struct mailbox` and the `class mailbox`. IMO it would be easier to differentiate them by renaming them explicitly MailboxResponse and MailboxRequest.
> 
> There is also the file `CoE/Mailbox`, which only contains the class SDOMessage, I would suggest to rename the file `SDOMessage` instead of mailbox.

I disagree: you almost enver use both together and I prefer to use namespace for this purpose (which are designed for that). Note that the old master code needs to be moved in another PR.

simulation/ESC.h
leducp: Right

lib/src/CoE/Mailbox.cc
leducp: Of course :)

lib/slave/src/AbstractESC.cc
leducp: Indeed. Should be fixed now.

Rdk-T:
> > I find the naming around the mailbox a bit confusing.
> > We have a namespace mailbox, with then the namespaces request/response. In these namespaces we have the `struct mailbox` and the `class mailbox`. IMO it would be easier to differentiate them by renaming them explicitly MailboxResponse and MailboxRequest.
> > There is also the file `CoE/Mailbox`, which only contains the class SDOMessage, I would suggest to rename the file `SDOMessage` instead of mailbox.
> 
> I disagree: you almost never use both together and I prefer to use namespace for this purpose (which are designed for that). Note that the old master code needs to be moved in another PR.

So the file Mailbox will be split in two ?

leducp:
> > > I find the naming around the mailbox a bit confusing.
> > > We have a namespace mailbox, with then the namespaces request/response. In these namespaces we have the `struct mailbox` and the `class mailbox`. IMO it would be easier to differentiate them by renaming them explicitly MailboxResponse and MailboxRequest.
> > > There is also the file `CoE/Mailbox`, which only contains the class SDOMessage, I would suggest to rename the file `SDOMessage` instead of mailbox.
> > 
> > 
> > I disagree: you almost never use both together and I prefer to use namespace for this purpose (which are designed for that). Note that the old master code needs to be moved in another PR.
> 
> So the file Mailbox will be split in two ?

I would like to split the mailbox in 2xN ->
2 -> request/resposne
N -> services (CoE, FoE, EoE, VoE, and so on)

lib/include/kickcat/CoE/OD.h
```diff
@@ -155,6 +156,8 @@ namespace kickcat::CoE
     struct Entry    // ETG1000.5 6.1.4.2.1 Formal model
     {
         Entry() = default;
+        Entry(uint8_t subindex, uint16_t bitlen, uint16_t access,
```
Rdk-T, line 159: l154, warning useless `;`

lib/src/Mailbox.cc
leducp: There is: if the message cannot be written, the message is kept in the sending list. Note that the ESC implementation shall check if the write can be done or the result of the write access.

Rdk-T (approved):

PR #145, opened 2024-03-18
Select specific arch nuttx file based on arch chip.

Rdk-T:
Filtering only on NuttX OS to build the slave lib is not enough. Some files are dependent on the chip arch.

### DIFF
--- a/examples/slave/CMakeLists.txt
+++ b/examples/slave/CMakeLists.txt
@@ -1,2 +1,5 @@
 add_subdirectory(nuttx/lan9252)
-add_subdirectory(nuttx/xmc4800)
+
+if (NUTTX_ARCH_CHIP STREQUAL "xmc4")
+    add_subdirectory(nuttx/xmc4800)
+endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ if (BUILD_EXAMPLES)
   if (UNIX)
       add_subdirectory(examples/master)
   endif()
-  if (NUTTX)
+  if (NUTTX AND (NUTTX_ARCH_CHIP STREQUAL "xmc4"))
     add_subdirectory(examples/slave)
```
leducp, line 70: rename slave to have slave_xmc4800?
It seems that if(NUTTX) is useless since is is always defined if NUTTX_ARCH_CHIP is defined.

Rdk-T: Hmm I moved this condition into the slave/CMakeLists.txt

We also have an example of NuttX using easycat, with the only dependency being SPI driver of NuttX.

leducp (approved):

PR #151, opened 2024-04-04
Virtual socket

leducp:
Add a TapSocket, a virtual socket that use  a shared memory to allow the use of KickCAT in emulated mode without the need for administrator rights.


examples/master/ingenia_control/ingenia_control.cc
```diff
@@ -45,7 +46,7 @@ int main(int argc, char *argv[])
         red_interface_name = argv[2];
     }
 
-    auto socket_nominal = std::make_shared<Socket>();
+    auto socket_nominal = std::make_shared<TapSocket>();
```
Rdk-T, line 49: Reminder reset example before merge

examples/master/load_esi/load_esi.cc
```diff
@@ -1,10 +1,42 @@
...
 
 int main()
 {
+    TapSocket socketA(true);
+    TapSocket socketB;
+    TapSocket socketC;
+
+    socketA.open("yolo");
+    socketB.open("yolo");
+    //socketC.open("yolo");
```
Rdk-T, line 17: Clean up example

lib/include/kickcat/OS/SharedMemory.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+    public:
+        SharedMemory();
+        SharedMemory(SharedMemory const& shm) = delete;
+        SharedMemory& operator=(SharedMemory const& shm) = delete;
+        ~SharedMemory();
+
+        /// \param  name    Name of the shared memory.
+        /// \param  size    Size of the shared memory in bytes.
+        /// \param  address Address where the shml segment shall be mapped. AUtomatiuc address if nullptr
```
Rdk-T, line 21: AUtomatiuc typo ?

simulation/network_simulator.cc
```diff
@@ -27,10 +28,10 @@ int main(int argc, char* argv[])
 
     CoE::EsiParser parser;
     auto& coe_dict = CoE::dictionary();
-    coe_dict = parser.load("ingenia_esi.xml");
+    coe_dict = parser.load("evs-net-01_esi_2.2.3.xml");
```
Rdk-T, line 31: Is this file accessible somewhere ?

simulation/network_simulator.cc
```diff
@@ -1,4 +1,5 @@
 #include "kickcat/OS/Linux/Socket.h"
+#include "kickcat/TapSocket.h"
```
Rdk-T, line 2: Is there updates to put in the readme about how to use the simulator with TapSocket ?

simulation/network_simulator.cc
leducp: No, but it can be downloaded on this manufacturer website :)

On a side note: the whole mechanism shall be modified to better described all the slaves that run on the simulator (who mentioned an ENI file?)

simulation/network_simulator.cc
leducp: Not really, but I'll revert this change specifically (I want to do the switch when the pcap trace will bve available)

Rdk-T (approved):

PR #161, opened 2024-05-31
SDO access hook

leducp:


lib/src/CoE/mailbox/response.cc
```diff
@@ -212,21 +232,43 @@ namespace kickcat::mailbox::response
...
+            auto* entry = &object->entries.at(subindex);
+            if (not isDownloadAuthorized(entry))
             {
                 abort(CoE::SDO::abort::READ_ONLY_ACCESS);
                 return ProcessingResult::FINALIZE;
             }
 
-            uint32_t entry_size = entry.bitlen / 8;
-            std::memcpy(entry.data, payload_ + 4 + size, entry_size);
+            beforeHooks(entry);
```
nicolas71640, line 242: How will we know if the master ask for a upload or download ?

leducp: No knowledge of the access type at this point: do you think it is needed?

nicolas71640: Well, I think this the least we can tell to the caller : what the master trying to do, as these hooks are made to react on these "demands"

leducp: Fair enough: I'll add the command to the request

nicolas71640 (approved):

PR #164, opened 2024-09-12
Parse string(n) from esi

nicolas71640:
Add the parsing of STRING(n) from ESI XML

### DIFF
--- a/lib/src/CoE/OD.cc
+++ b/lib/src/CoE/OD.cc
@@ -146,7 +146,6 @@ namespace kickcat::CoE
             case CoE::DataType::VISIBLE_STRING:
             {
 				std::string strResult{static_cast<char const *>(data)};
-				reverse(strResult.begin(), strResult.end());
 				return strResult;
             }
             default: {}
--- a/lib/src/CoE/OD.cc
+++ b/lib/src/CoE/OD.cc
@@ -1,6 +1,7 @@
 #include <sstream>
 #include <algorithm>
 #include <streambuf>
+#include <string_view>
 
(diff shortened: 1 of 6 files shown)


leducp (approved):

leducp (approved):

lib/src/CoE/OD.cc
```diff
@@ -142,6 +143,12 @@ namespace kickcat::CoE
                 result << *static_cast<int64_t const *>(data);
                 break;
             }
+            case CoE::DataType::VISIBLE_STRING:
+            {
+				std::string strResult{static_cast<char const *>(data)};
```
leducp, line 148: alignment

lib/src/CoE/OD.cc
```diff
@@ -306,4 +313,13 @@ namespace kickcat::CoE
         static Dictionary dict;
         return dict;
     }
+ 
+ 	void addEntryString(Object &object, uint8_t subindex, uint16_t bitlen, uint16_t access,
```
leducp, line 317: alignment

lib/src/CoE/OD.cc
```diff
@@ -306,4 +313,13 @@ namespace kickcat::CoE
...
+              DataType type, std::string const& description, std::string data)
+    {
+        object.entries.emplace_back(subindex,bitlen,access,type,description);
+        std::string* dataAlloc = new std::string(data);
```
leducp, line 321: Copy the content instead of the object to avoid leak if the entry is deleted (through free, not delete)

PR #165, opened 2024-09-14
Fix Download response

nicolas71640:
In the download response the command wasn't set (for download and segmented download). 

Tested for download, not for segmented download, but I'm no worry for this last one.

Edited : I actually also add the command se in the upload and segmented upload. I suppose it was working before because the request::UPLOAD and response::UPLOAD was both equal to 2, and that the slave stack has been used to do segmented upload.

### DIFF
--- a/lib/src/CoE/mailbox/response.cc
+++ b/lib/src/CoE/mailbox/response.cc
@@ -200,6 +200,7 @@ namespace kickcat::mailbox::response
         std::memcpy(entry->data, payload_, size);
 
         coe_->service = CoE::Service::SDO_RESPONSE;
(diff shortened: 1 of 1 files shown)


leducp (approved):

leducp:
> In the download response the command wasn't set (for download and segmented download).
> 
> Tested for download, not for segmented download, but I'm no worry for this last one.
> 
> Edited : I actually also add the command se in the upload and segmented upload. I suppose it was working before because the request::UPLOAD and response::UPLOAD was both equal to 2, and that the slave stack has been used to do segmented upload.

Nice catch!

PR #166, opened 2024-10-11
Inject dictionary instead of singleton

nicolas71640:
Due to simulation purpose, the injection of the CoE::dictionary is more appropriate.


lib/include/kickcat/CoE/OD.h
```diff
@@ -220,7 +221,7 @@ namespace kickcat::CoE
         }
     }
 
-    void populateOD();
+    std::shared_ptr<Dictionary> createOD();
 
     // Singleton
```
Rdk-T, line 226: Remove singleton comment ?

Rdk-T (approved):

leducp (approved):

PR #167, opened 2024-10-28
Make os_types.h private

nicolas71640:


lib/include/kickcat/types.h
```diff
@@ -0,0 +1,10 @@
+#ifndef KICKCAT_TYPES_H
+#define KICKCAT_TYPES_H
+
+#if defined(__linux__) || defined(__NuttX__)
```
leducp, line 4: __unix__ ? (which means that we may also create a directory Unix that is common between Linux and NuttX)

leducp (approved):

PR #180, opened 2024-11-13
Fix eeprom end mark sometimes not catched

### LINKED ISSUE #152: [master] Read eeprom never ends
I have used the eeprom a the end of this post with the simulator (I think this not related to the simulator however), and the eeprom fetching never ends. After investigation with @Rdk-T we've figured that this line was the culprit : 
                return (((s->sii.buffer.back() >> 16) == eeprom::Category::End) and (s->sii.buffer.size() > 32));

The last word of the eeprom is 0x0000FFFF, the 16bits shift gives 0x0000 so the end half word is not seen as it should be. 

Here's the faulty_eeprom to reproduce the issue

[faulty_eeprom.zip](https://github.com/leducp/KickCAT/files/15022261/faulty_eeprom.zip)

leducp:
close #152

### DIFF
--- a/lib/slave/src/ESC/EmulatedESC.cc
+++ b/lib/slave/src/ESC/EmulatedESC.cc
@@ -360,21 +360,37 @@ namespace kickcat
         {
             case eeprom::Control::READ:
             {
-                std::memcpy((void*)&memory_.eeprom_data, eeprom_.data() + memory_.eeprom_address, 4);
+                if (memory_.eeprom_address < eeprom_.size())
+                {
+                    std::memcpy((void*)&memory_.eeprom_data, eeprom_.data() + memory_.eeprom_address, 4);
+                }
+                else
+                {
+                    // Eeprom address is out of bound: return the default value of an unwritten eeprom.
+                    // This is a shortcut for emulation, real device should handle the eeprom size.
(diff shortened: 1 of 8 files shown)


leducp:
Tested on wdc_foot with the example and the eeprom tool (dump

nicolas71640 (approved):

lib/slave/src/ESC/EmulatedESC.cc
```diff
@@ -360,21 +360,37 @@ namespace kickcat
...
-                std::memcpy((void*)&memory_.eeprom_data, eeprom_.data() + memory_.eeprom_address, 4);
+                if (memory_.eeprom_address < eeprom_.size())
+                {
+                    std::memcpy((void*)&memory_.eeprom_data, eeprom_.data() + memory_.eeprom_address, 4);
+                }
+                else
+                {
+                    // Eeprom address is out of bound: return the default value of an unwritten eeprom.
+                    // This is a shortcut for emulation, real device should handle the eeprom size.
+                    //memory_.eeprom_data = UINT64_MAX;
```
Rdk-T, line 371: To fix

Rdk-T (approved):

PR #183, opened 2024-11-22
Windows support

### LINKED ISSUE #33: Add helper to display available interfaces
In order to have more user friendly example it would be nice to display the available interfaces.

Thanks.

### LINKED ISSUE #96: Windows support
Add a windows socket and a windows build system through CMake

leducp:
close #96 
close #33 

Add Windows support through WinPCAP API (npcap on latest Windows versions).


conan/conanfile_windows.txt
```diff
@@ -0,0 +1,10 @@
+[requires]
+gtest/1.15.0
+tinyxml2/10.0.0
+npcap/1.70
```
trns1997, line 4: Is there a reason why you chose `npcap` instead of `wincap`? Better alternative? More actively maintained? Just curious

conan/profile_windows_x86_64.txt
```diff
@@ -0,0 +1,16 @@
+[settings]
+os=Windows
+arch=x86_64
+compiler=gcc
```
trns1997, line 4: You can GCC to compile on windows???

leducp: Of course: MinGW is GCC for Windows. And I don't want to messup with MSVC compiler since we heavily rely on some not that really standards features (likes bit fields).

trns1997: Ah good to know, I always though MinGW and GCC were 2 different things. Of course passing by GCC simplifies compatibility.

conan/conanfile_windows.txt
leducp: WinPCAP is dead: it is not maintained and it rely on the NDIS 5.0 API, which was superseed by the 6.0 version in Windows Vista.

The 5.0 API is still eanbled in the latest WIndows version as of today, but it is mark legacy and could be removed anytime (especially since it seems that there is a potential cybersecurity breach when using it).

However, npcap uses the winpcap API, so one can just link with winpcap and it should work.

So to summarize:
WinPCAP pros:
* free
* available on all systems (since Windows 98 to Windows 11)
WinPCAP cons:
* slow
* not maintained
* can be disabled anytime in the latest Windows version

NPCAP pros:
* fast
* maintained

NPCAP cons:
* not that free (but you are granted a few free licence if you do no use it for commercial purpose)

Rdk-T, line 4: Version mismatch with readme ?

Rdk-T (approved):

conan/conanfile_windows.txt
leducp: Because the driver installed for the test was the 1.80 (with the npcap 1.70 SDK). Overall it is working with this configuration :)

leducp: README updated to reflect the difference between the SDK (conan) and the driver (to be installed by the user)

Issue #206, opened 2025-08-13
Write EEPROM: check EEPROM Algorithm is compatible with file size


leducp:
Register 0x502:7
0: 1 address byte (1Kbit – 16Kbit EEPROMs)     -> max file is 2KB
1: 2 address bytes (32Kbit – 4 Mbit EEPROMs) -> max file is 512KB

If a file is too big, the eeprom write is invalid (even the first part is reset to 0xFF in the case of lan9254 + 24LC16B)

PR #207, opened 2025-08-14
eeprom tool update

leducp:
Fix: #206


tools/eeprom.cc
```diff
@@ -121,24 +153,43 @@ int main(int argc, char* argv[])
...
         std::ofstream f(file, std::ofstream::binary);
 
-        // Create an eeprom binary file with the right size of empty data
-        std::vector<char> empty(slave.sii.eeprom_size, -1);
-        f.write(empty.data(), empty.size());
-        f.seekp(0);
-
         // Write dumped data
-        f.write(raw_data, sii.size() * 4);
+        f.write(raw_data, slave.sii.eeprom.size() * 4);
```
matthieuvigne: Magic number. `sizeof(uint32_t)` ?

tools/eeprom.cc
```diff
@@ -153,6 +204,24 @@ int main(int argc, char* argv[])
         eeprom_file.read((char*)buffer.data(), size);
         eeprom_file.close();
 
+        if (addressingScheme2Bytes == false)
+        {
+            // max eeprom size is 16Kb
+            if ((buffer.size() * 2) > 2_KiB)
```
matthieuvigne: Clarify comment: to go from 16Kb to 2_KiB, I guess it's because we write 32 bits words. But why buffer.size() * 2 ?

matthieuvigne (approved):

tools/eeprom.cc
leducp: Right: I switched to the size variable a few line before to skip useless computation and have a direct size mapping on bytes.

trns1997: buffer.size() seems to be the number of works and the `* 2` converts words to byte. (Hopefully i am correct in saying this)

tools/eeprom.cc
```diff
@@ -147,12 +198,30 @@ int main(int argc, char* argv[])
...
+                printf("!!! WARNING !!!\n");
+                printf("Current EEPROM addressing scheme is one byte: max EEPROM size is 16Kbit (2KiB)\n");
+                printf("but the given EEPROM file size is %ldB which exceed the maximum possible target size\n", size);
+                printf("If you wish to continue, the written size will be truncated to 2KiB\n");
+
+                if (not askContinue())
+                {
+                    return 0;
+                }
+                buffer.resize(2_KiB / 2);
```
trns1997, line 221: why truncate here and not just refuse the write?

leducp: Well I asked the user: if he wish to continue then the buffer shall at least be truncated.Not perfect and we could just refuse to do anything but it seems usefull when the useful part of the given file is small enough.

tools/eeprom.cc
```diff
@@ -21,6 +21,32 @@
...
 
+bool askContinue()
+{
+    while (true)
+    {
+        printf("Do you want to continue? (y/n): ");
+
+        char input[16];
+        char answer;
+        if (fgets(input, sizeof(input), stdin) != nullptr)
```
trns1997, line 32: any reason why you do not use cin and string like:

```
bool askContinue()
{
    std::string input;
    while (true)
    {
        std::cout << "Do you want to continue? (y/n): ";
        if (!std::getline(std::cin, input))
            continue;

        if (!input.empty())
        {
            char c = input[0];
            if (c == 'y' || c == 'Y') return true;
            if (c == 'n' || c == 'N') return false;
        }
    }
}
```

(Not tested)

leducp: To stay coherent with the use of printf to manage standard inputs/outputs. No bigger reason.

PR #210, opened 2025-08-15
slight rework of the spi interface

### LINKED ISSUE #209: Rework SPI class
Add setters and getters to pass CPOL/CPH modes, baudrate and other such attributes to have a less hardcoded class usage

trns1997:
Closes #209

- [ ] Tested
- [x] Update SPIMock

### DIFF
--- a/lib/slave/driver/include/kickcat/nuttx/SPI.h
+++ b/lib/slave/driver/include/kickcat/nuttx/SPI.h
@@ -1,29 +1,46 @@
-
 #ifndef SLAVE_STACK_INCLUDE_NUTTX_SPI_H_
 #define SLAVE_STACK_INCLUDE_NUTTX_SPI_H_
 
 #include "kickcat/AbstractSPI.h"
 #include <nuttx/spi/spi_transfer.h>
-
-#include <stdio.h>
+#include <string>
 
 namespace kickcat
 {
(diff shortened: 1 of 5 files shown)


lib/slave/driver/include/kickcat/nuttx/SPI.h
```diff
@@ -1,29 +1,46 @@
...
+
+        void setChipSelect(uint32_t cs) override { chipSelect_ = cs; }
+        uint32_t getChipSelect() const override { return chipSelect_; }
 
     private:
         file filep_spi_;
         int fd_;
         spi_dev_s** priv_spi_;
+
+        std::string devicePath_;
```
trns1997, line 39: Not sure to what extend we can move these attributes to the abstract class? They are more or less common to all spi interfaces.

leducp: baudrate can be moved as is, but mode need to have an enum to abstract it from the underlying API.

trns1997: In that case let's just keep the abstract class as the class blueprint and as new child classes are created we'll see what can be moved to the abstract class.

leducp:
superseed by #220

PR #211, opened 2025-08-20
Quality of life - network interface selector

leducp:
Add a helper to select the network interface in the examples/tools.

### DIFF
--- a/lib/master/src/helpers.cc
+++ b/lib/master/src/helpers.cc
@@ -6,6 +6,8 @@
 #include "kickcat/SocketNull.h"
 #ifdef __linux__
 #include "kickcat/OS/Linux/Socket.h"
+#elif __PikeOS__
+    #include "kickcat/OS/PikeOS/Socket.h"
 #elif __MINGW64__
 #include "kickcat/OS/Windows/Socket.h"
 #else
--- a/lib/master/src/helpers.cc
+++ b/lib/master/src/helpers.cc
@@ -1,7 +1,16 @@
 #include <iostream>
 
 #include "helpers.h"
-#include "AbstractSocket.h"
+
+#include "kickcat/TapSocket.h"
+#include "kickcat/SocketNull.h"
+#ifdef __linux__
+#include "kickcat/OS/Linux/Socket.h"
+#elif __MINGW64__
+#include "kickcat/OS/Windows/Socket.h"
+#else
+#error "Unsupported platform"
+#endif
 
 namespace kickcat
 {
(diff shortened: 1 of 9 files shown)


lib/master/src/helpers.cc
```diff
@@ -23,23 +34,80 @@ namespace kickcat
             return; // No interactive choice asked: nothing to do
         }
 
-        auto netIntefaces = listInterfaces();
-        for (std::size_t i = 0; i < netIntefaces.size(); ++i)
+        auto net_interfaces = listInterfaces();
```
trns1997, line 39: I feel the code style is starting to get inconsistent. I suggest we add a code format checker like `clang-format` what do you guys think? I create an issue if you guys think it's a good idea.

leducp: It is (the project started to be too big with too many peoples so yes, definitly et good idea)!  

But I'm not sure where you find it inconsistent? (here I clean the variable style to snake_cases to follow the style)

trns1997: Ah no the file here is fine its just my PR before was in camelCase where as here we have variable's in snake_case. I'll look into harmonizing all this across the project without breaking anything hopefully :/

simulation/network_simulator.cc
```diff
@@ -55,11 +56,11 @@ int main(int argc, char* argv[])
     }
 
     CoE::EsiParser parser;
-    auto coe_dict = parser.loadFile("wdc_foot_eve_beta.xml");
+    auto coe_dict = parser.loadFile("foot.xml");
 
     printf("Start EtherCAT network simulator on %s with %ld slaves\n", argv[1], escs.size());
-//    auto socket = std::make_shared<TapSocket>(true);
-    auto socket = std::make_shared<Socket>();
+    auto socket = std::make_shared<TapSocket>(true);
```
trns1997, line 61: Why is this now a tap socket?

leducp: It depends on what the simulator should work on: local/virtual or real network.

However this section of the code is still POC: the simulator is not really configurable : network descriptors, fake slaves, network, everything is hardcoded for now.

lib/src/OS/NuttX/Socket.cc
trns1997, line 1: Is there a reason for a NuttX socket implementation, even though NuttX is largely POSIX-compliant. Ccan we not reuse the Linux (or a generic Unix) socket implementation for both? I am guesing it is mainly because of `listInterfaces` but maybe in the future when we have interface discovery on nuttx these 2 can be fused to a Unix socket?

lib/include/kickcat/OS/Windows/types/os_types.h
```diff
@@ -0,0 +1,26 @@
+#ifndef KICKCAT_OS_WINDOWS_OS_TYPES_H
+#define KICKCAT_OS_WINDOWS_OS_TYPES_H
+
+#define WIN32_LEAN_AND_MEAN
```
trns1997, line 4: Hahah can we rename this to `WIN32_LEAN`

lib/src/OS/Windows/SharedMemory.cc
```diff
@@ -0,0 +1,69 @@
...
+            INVALID_HANDLE_VALUE,  // use system paging file
+            nullptr,               // default security
+            PAGE_READWRITE,        // read/write access
+            static_cast<DWORD>((size_ >> 32) & 0xFFFFFFFF),  // high-order size
+            static_cast<DWORD>(size_ & 0xFFFFFFFF),          // low-order size
+            name.c_str());         // name of the mapping object
+
+        if (fd_ == nullptr)
+        {
+            throw std::runtime_error("CreateFileMapping() failed");
```
trns1997: use `THROW_LAST_ERROR` to be consistent across the file?

lib/src/OS/Windows/SharedMemory.cc
```diff
@@ -0,0 +1,69 @@
...
+    {
+    }
+
+    SharedMemory::~SharedMemory()
+    {
+        if (address_ != nullptr)
+        {
+            if (not UnmapViewOfFile(address_))
+            {
+                THROW_LAST_ERROR("UnmapViewOfFile() failed");
```
trns1997, line 23: do you really want to throw in destruction? Maybe just logging might be better, functionally it has no impact just the clean up failed we do not want the application to throw for this non?

lib/src/OS/NuttX/Socket.cc
leducp: There is no direct equivalent: sure it is POSIX but the Linux version uses Linux specific API to handle the kernel specific behavior/API (like IRQ coalescing and timeout management).

lib/include/kickcat/OS/Windows/types/os_types.h
leducp: Well, you  have to ask Microsoft :rofl:

lib/src/OS/Windows/SharedMemory.cc
leducp: Indeed!

lib/src/OS/Windows/SharedMemory.cc
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+    }
+
+    void SharedMemory::open(std::string const& name, std::size_t size, void* address)
+    {
+        size_ = size;
+
+        // Create or open shared memory (file mapping in system paging file)
+        fd_ = CreateFileMappingA(
```
trns1997, line 41: I see a lot of people on stackoverflow using `CreateFileMappingW`, it seems to be the same thing. Here https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga vs https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-createfilemappingw
but i do not really know the subtleties. Just curious whats the difference ;)

lib/src/OS/Windows/SharedMemory.cc
leducp: Well it is already the strategy in the Unix variant and this is the only means to ensure that the application detect a problem which is OS related. In this situation, there is no real solution: something goes wrong so you HAVE to acknowledge it. Note that shared memories are only in use for the tap socket so destruction is done at the end of the software (simulator usage).

lib/src/OS/Windows/SharedMemory.cc
leducp: If I understood correctly (but keep in mind that if I'm experienced in POSIX API, I'm a total beginner with Win32 API), the A variant stands for ANSI and W for Wide -> it defines the encoding used by the string lpName in the documentation (A -> ANSI encoding, C-style, W -> Wide -> UTF-16)

PR #225, opened 2025-11-28
Add minimal Python bindings

### LINKED ISSUE #229: Python binding
Add minimal Python binding

leducp:
close #229


lib/py_bindings/src/master_py.cc
```diff
@@ -0,0 +1,187 @@
...
+{
+    void create_master_python_bindings(nb::module_ &m)
+    {
+        nb::enum_<State>(m, "State")
+            .value("INIT", State::INIT)
+            .value("PREOP", State::PRE_OP)
+            .value("SAFE_OP", State::SAFE_OP)
+            .value("OPERATIONAL", State::OPERATIONAL);
+
+        nb::class_<Slave>(m, "Slave")
```
trns1997: I was thinking maybe later split these into seperate files like `slave_py.cc`, `link_py.cc` and so on so we can bind other classes as we go instead of having 1 massive file?

lib/py_bindings/src/bus_py.cc
```diff
@@ -0,0 +1,187 @@
...
+        m.def("create_link", [](std::string const& nom, std::string const& red)
+            {
+                auto [s1, s2] = createSockets(nom, red);
+                return std::make_shared<Link>(s1, s2, []()
+                {
+                    printf("Redundancy activated\n");
+                });
+            });
+
+        nb::class_<Bus>(m, "Bus")
```
trns1997, line 32: yes i think we should consider splitting these into different files, this should go into `bus_py.cc` what do you think?

py_bindings/examples/easycat.py
```diff
@@ -0,0 +1,49 @@
...
+def cyclic_callback():
+    bus.process_data()
+
+bus.wait_for_state(State.OPERATIONAL, 1.0, cyclic_callback)
+
+print("After OPERATIONAL - Slave info:")
+for s in bus.slaves():
+    print(f" - Slave {s.address} input: {s.input_size} output: {s.output_size}")
+
+easycat = bus.slaves()[0]
```
trns1997: ah not sure you need this one? I had problems initially where the `input_data` was always empty, i thought it was because my reference to the slave was out dated. I think now that we call `bus.create_mapping()` before fetching the slave reference, this is no longer required.

trns1997: I recommend testing to be sure

leducp: It is not needed per see but it show how to manipulate a slave no?

trns1997: We already have a reference to the slave in line 19

leducp: right: removed (I tested it is OK)

Issue #201, opened 2025-05-23
Imrpove coverage


leducp:
Acceptable values:
* >95% line
* >80% branches

PR #241, opened 2025-12-12
Improve coverage

leducp:
related to #201

### DIFF
--- a/unit/src/slave/Slave-t.cc
+++ b/unit/src/slave/Slave-t.cc
@@ -1,0 +1,68 @@
+#include "mocks/ESC.h"
+
+#include "kickcat/slave/Slave.h"
+
+using namespace kickcat;
+using namespace kickcat::slave;
+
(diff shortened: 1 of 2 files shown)


unit/src/slave/Slave-t.cc
```diff
@@ -0,0 +1,68 @@
...
+};
+
+TEST_F(SlaveTest, routine_without_mailbox)
+{
+    slave_.start();
+    slave_.routine();
+    ASSERT_EQ(slave_.state(), State::INIT);
+}
+
+TEST_F(SlaveTest, routine_with_mailbox)
```
trns1997, line 48: hmmm i do not get the tests, they work but i feel like they don't test much, maybe we should add a counter example (maybe in an other PR) where we show the slave ends up in a problem state, to cover all branches

leducp: agree
Put it in draft for the moment

PR #245, opened 2025-12-13
Ethercat gui

### LINKED ISSUE #264: [GUI Tool] PDO Interaction Panel (Read)
Provide a GUI panel to interact with PDO data once the slave reaches SAFE-OP or OP.

Features:
- Display mapped PDO entries
- Read live PDO values

### LINKED ISSUE #266: [GUI Tool] Get Bus and Slave infos
Get bus info and slave infos. 

- List slave devices on the bus
- Select slave on the bus
- Display slave state on the bus
- Transition slave on the bus

### LINKED ISSUE #267: [GUI Tool] Read and Display Slave dataframe contents

trns1997:
Closes #267 
Closes #266 
Closes #264 

[Screencast from 2025-12-13 22-12-59.webm](https://github.com/user-attachments/assets/21e12c42-23a9-456e-8acc-4993ed5f7215)

This is a very crude POC to see if a gui application is doable. If we want to take this idea further, we will need to rethink the archi (there is none), create clear components based on what we want.

### Usage:
```
pip install PySide6
pip install . # for the latest binding capabilities
python -m tools.ethercat_gui -i enp8s0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build*/
 compile_commands.json
 .clangd
 flashBin.jlink
+__pycache__/
--- a/README.md
+++ b/README.md
@@ -457,6 +457,7 @@ KickCAT includes several utility tools in the `tools/` directory:
 
 - **EEPROM tools**: Read/write/dump EEPROM from ESC
 - **OD Generator**: Generate Object Dictionary code from ESI files
+- **EtherCAT GUI**: PySide6-based graphical interface for bus monitoring
 
 Build tools with the main project:
 ```bash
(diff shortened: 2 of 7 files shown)


trns1997:
> I like the general idea, especially to ease PDO mapping. I would proably use it if I can setup the PDO mapping manullay through the GUI. The flow I have in mind is like this:
> 
>     1. add/rm a few SDO to the mapping (with a button 'add' and a field id/subid)
> 
>     2. apply the mapping and check if it was written successfully
> 
>     3. try the mapping (because writing OK doesn't mean that the mapping is coherent/valide);  if failure, go back to 1.
> 
>     4. go to SAFE-OP;  if failure, go back to 1
> 
>     5. read/write PDO
> 
> 
> Another thing that could be nice is to have a standard GUI for DS402 slaves. So you can select the slave (no need to detect the profile for a starter), go to OP (the app apply a standard PDO configuration - we probably need something configurable with a file to handle various hardware) then enable/disable the motor, apply a torque, go to pos, etc

I rearchitectured and split up the code to make it cleaner to build upon. So i suggest we validate this GUI tool as it is first. And then i will make a bunch on issues based on your requested features, to build them piece by piece. What do you think?

leducp (approved):

PR #251, opened 2025-12-17
fix config to have running xmc4 slave and better UX easycat py

trns1997:


lib/py_bindings/src/bus_py.cc
```diff
@@ -72,6 +72,12 @@ namespace kickcat
                     self.processDataRead(read_error);
                     self.processDataWrite(write_error);
                 })
+            .def("process_data_noop", [](Bus &self)
```
trns1997: do we want to rework? This is essentially process_data() but without the datagram state. Not sure if it is normal that we need this. But without it the xmc fails to cycle through

trns1997: ping @leducp

leducp: It is required to have a valid PDO communication in SAFEOP otherwise the slave should refuse to go in OP.
However part of the communicaiton may not be acknowledged by the slave because SAFEOP varies from salve to slave (observed behavior, dunno why theses changes)

leducp: I think it is good to have an explicit no_error version in Python to ease the tools/examples.
Can you rename it liike this? (something like: process_data_no_check)

trns1997: I can rename it, I was thinking of passing python lambda in the param so that we only have 1 process_data. But I find it makes use api a bit more complex. So I think your idea of keeping it simple for the user is a good idea. We can change this if in the future we want to mimic perfectly the c++ api :)

PR #255, opened 2025-12-18
mailbox necessary for preop to safeof transition

### LINKED ISSUE #252: xmc4800-relax unable to transitiion to safe op

trns1997:
Closes #252

### DIFF
--- a/lib/slave/src/AbstractEmulatedEEPROM.cc
+++ b/lib/slave/src/AbstractEmulatedEEPROM.cc
@@ -136,12 +136,6 @@ namespace kickcat
             printf("Checksum Mismatch !! Load default eeprom \n");
             load_default_eeprom_ = true;
         }
-
-        // debug print:
-        for(uint32_t i = 0; i < 20; i++)
-        {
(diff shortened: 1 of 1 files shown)


trns1997:
@leducp merge this before #251

lib/slave/src/AbstractEmulatedEEPROM.cc
```diff
@@ -129,19 +129,10 @@ namespace kickcat
 
         uint32_t adler_sum = adler32Sum(eeprom_.data(), eeprom_.size() * 2);
 
-        printf("Adler sum read %" PRIx32 ", computed %" PRIx32 "\n", adler_checksum_, adler_sum);
```
leducp, line 132: why the change?

trns1997: these are debug prints they do not seem necessary. Do you wish to keep this?

leducp: I propose to keep the trace mismatch and add inside the computed and expected checksum and to add another trace "OK" if everything is fine. 

This is a debug trace but it is nice ti have them when your board starts to fail (flash corruption for instance)

PR #257, opened 2025-12-19
dynamic pdo configuration slave

### LINKED ISSUE #226: [Slave] Support automatic PDO configuration + contiguous PDO memory mapping
Add automatic PDO configuration support: read 0x1C12/0x1C13, follow 0x1600/0x1A00 mappings, build a contiguous in-memory PDO buffer, and provide helpers to copy that buffer to/from the ESC process-data and to pack/unpack into a C struct.

trns1997:
Closes #226 

### Slave:
1. Build and deploy:
```
./scripts/build_slave_bin.sh freedom-k64f ~/nuttxspace/nuttx && ./examples/slave/nuttx/lan9252/freedom-k64f/deploy.sh build_freedom-k64f/easycat_frdm_k64f.bin
```

### Master
1. Static map example
```
sudo ./examples/master/freedom-k64f/freedom_k64f_static_map_example enp8s0
```

2. Master requests only accel example
```
sudo ./examples/master/freedom-k64f/freedom_k64f_accel_map_example enp8s0
```

3. Master requests only mag example
```
sudo ./examples/master/freedom-k64f/freedom_k64f_mag_map_example enp8s0
```

## TODO:
- [x] Move mapping api lower level make it transparent for user
<del> - [ ] Add TU for new functionalities </del> Will do it in a seperate PR: https://github.com/leducp/KickCAT/issues/278

### DIFF
--- a/examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
+++ b/examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
@@ -65,7 +65,7 @@ namespace kickcat::CoE
             CoE::addEntry<uint32_t>(object,3,32,72,CoE::Access::READ | CoE::Access::WRITE,CoE::DataType::UNSIGNED32,"TxPDO Map 1 Element 3",0x60020010);
             CoE::addEntry<uint32_t>(object,4,32,104,CoE::Access::READ | CoE::Access::WRITE,CoE::DataType::UNSIGNED32,"TxPDO Map 1 Element 4",0x60030010);
             CoE::addEntry<uint32_t>(object,5,32,136,CoE::Access::READ | CoE::Access::WRITE,CoE::DataType::UNSIGNED32,"TxPDO Map 1 Element 5",0x60040010);
-            CoE::addEntry<uint32_t>(object,6,32,168,CoE::Access::READ,CoE::DataType::UNSIGNED32,"TxPDO Map 1 Element 6",0x60050010);
+            CoE::addEntry<uint32_t>(object,6,32,168,CoE::Access::READ | CoE::Access::WRITE,CoE::DataType::UNSIGNED32,"TxPDO Map 1 Element 6",0x60050010);
             dictionary.push_back(std::move(object));
         }
 


examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
```diff
@@ -0,0 +1,265 @@
...
+        {
+            State state = bus.getCurrentState(slave);
+            printf("Slave %d state is %s\n", slave.address, toString(state));
+        }
+    };
+
+    uint8_t io_buffer[2048];
+
+    // Copied from Ingenia Control example
+    const auto mapPDO = [&](const uint8_t slaveId, const uint16_t PDO_map, const uint32_t *data, const uint32_t dataSize, const uint32_t SM_map) -> void
```
trns1997: Should we make this a global method that can be used by user to remap/reconfigure?

leducp: Seems to be a good idea to create a generic helper. However we need to add an emulated version of complete access to do so: ingenia is quite the exception to support this option.
I already wrote an emulated compelte access for read that can be serve as an example, and the API is ready (better: read and write will be similar).

trns1997: I’m not entirely sure I understand your point. I copied this lambda from the Ingenia example. In this code, Complete Access is set to false, which I assumed meant the SDO write is effectively using an emulated CA path. From my understanding, for the Frdm case we could set CA to true, which would be more representative of actual Complete Access support (not sure if I understand this correctly).
Also, why does writeSDO take Complete Access as a boolean instead of using the Access enum? Is this just a leftover from an older refactor?

leducp: It is not? (line 113).
Here after a draft PR to cleanup the API:
https://github.com/leducp/KickCAT/pull/260

I remembered now why I didn't implement a writeSDO with "emulated complete access"  before: when writing, you have to know what you are targeting (array, var, object) and the underlying type (i.e. an array of u16, and first index is always u8 for array).

However I modified the ingenia example to switch back to "emulated complete access" (possible because we know the targeted type)

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,89 @@ int main(int argc, char *argv[])
...
+                *ay = sensor_data.accel.y;
+                *az = sensor_data.accel.z;
+                *mx = sensor_data.magn.x;
+                *my = sensor_data.magn.y;
+                *mz = sensor_data.magn.z;
+            }
+
+            // Update LEDs based on output PDO
+            userled_set_t led_set = 0;
+            if (*led_r)
```
leducp, line 146: {} around if

leducp: why do you need to chec the pointer? There will always be a led, even if the master don't ask for it.

examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
```diff
@@ -59,13 +60,15 @@ namespace kickcat::CoE
...
                 {}
             };
-            CoE::addEntry<uint8_t>(object,0,8,0,7,static_cast<CoE::DataType>(5),"Subindex 000",0x6);
-            CoE::addEntry<uint32_t>(object,1,32,16,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 1",0x60000010);
-            CoE::addEntry<uint32_t>(object,2,32,48,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 2",0x60010010);
-            CoE::addEntry<uint32_t>(object,3,32,80,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 3",0x60020010);
-            CoE::addEntry<uint32_t>(object,4,32,112,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 4",0x60030010);
-            CoE::addEntry<uint32_t>(object,5,32,144,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 5",0x60040010);
-            CoE::addEntry<uint32_t>(object,6,32,176,7,static_cast<CoE::DataType>(7),"TxPDO Map 1 Element 6",0x60050010);
+            // CHANGED: Access from 7 (read-only) to 63 (read-write in PRE_OP)
```
leducp: useless comment for merge

examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
```diff
@@ -44,10 +44,11 @@ namespace kickcat::CoE
                 "RxPDO Map 1",
                 {}
             };
-            CoE::addEntry<uint8_t>(object,0,8,0,7,static_cast<CoE::DataType>(5),"Subindex 000",0x3);
-            CoE::addEntry<uint32_t>(object,1,32,16,7,static_cast<CoE::DataType>(7),"RxPDO Map 1 Element 1",0x70000008);
-            CoE::addEntry<uint32_t>(object,2,32,16,7,static_cast<CoE::DataType>(7),"RxPDO Map 1 Element 2",0x70010008);
-            CoE::addEntry<uint32_t>(object,3,32,16,7,static_cast<CoE::DataType>(7),"RxPDO Map 1 Element 3",0x70020008);
+            // CHANGED: Access from 7 (read-only) to 63 (read-write in PRE_OP)
```
leducp: useless comment

lib/slave/src/PDO.cc
```diff
@@ -65,25 +65,156 @@ namespace kickcat
 
     void PDO::updateInput()
     {
-        if (input_ != nullptr)
+        if (not input_)
```
leducp: prefer "== nullptr"

lib/slave/include/kickcat/PDO.h
```diff
@@ -23,7 +24,24 @@ namespace kickcat
...
+            void*    od_data;   // pointer to OD entry->data
+            uint16_t bit_len;   // size in bits
+            uint16_t bit_offset;   // offset in PDO buffer (bits)
+        };
+
+        std::vector<uint16_t> parseAssignment(CoE::Dictionary& dict, uint16_t assign_idx);
+
+        bool parsePdoMap(CoE::Dictionary& dict, uint16_t pdo_idx, std::vector<PdoEntry>& map, uint16_t& bit_offset);
+
+        std::vector<PdoEntry> input_map_;
```
leducp: It is better to modifiy the dictionary pointers directly to avoid a copy in the updateInput() and updateOutput().
Something like this:

```
auto [index, subindex, size] = parseNextEntry(0x1c12);
auto [object, entry] = CoE::findObject(dict, index, subindex);
if (size != entry->bitlen)
{
    reportErrorToMaster(AbortCode::WRONG_SIZE)
}
void* save_ptr = entry->data;
entry->data = input_ + current_pos;
current_pos += size;
std::memcpy(entry->data, save_ptr);
free save_ptr;
```

lib/slave/src/PDO.cc
```diff
@@ -65,25 +65,156 @@ namespace kickcat
...
+            return;
+        }
+
+        std::memset(input_, 0, sm_input_.length);
+
+        for (auto const& e : input_map_)
+        {
+            uint8_t* dst = static_cast<uint8_t*>(input_) + (e.bit_offset / 8);
+
+            std::memcpy(dst, e.od_data, e.bit_len / 8);
```
leducp: useless copy if the dictionary pointer is relinked (cf. previous comment)

lib/slave/src/PDO.cc
```diff
@@ -65,25 +65,156 @@ namespace kickcat
...
+
+        uint8_t count = *static_cast<uint8_t*>(entry0->data);
+
+        for (uint8_t i = 1; i <= count; ++i)
+        {
+            auto [obj, entry] = CoE::findObject(dict, pdo_idx, i);
+            if (not entry)
             {
-                slave_error("\n update_process_data_output ERROR\n");
+                return false;
```
leducp, line 137: better to return a corresponding errno (like ENOENT here)

lib/slave/src/PDO.cc
```diff
@@ -65,25 +65,156 @@ namespace kickcat
...
+        input_map_.clear();
+        output_map_.clear();
+
+        {
+            uint16_t bit_offset = 0;
+            std::vector<uint16_t> pdo_indices = parseAssignment(dict, 0x1C13);
+
+            if (pdo_indices.empty())
+            {
+                return StatusCode::INVALID_INPUT_CONFIGURATION;
```
leducp: Maybe we can authorize an empty mapping no?

lib/slave/src/PDO.cc
leducp: Or better: directly the StatusCode?

lib/slave/src/PDO.cc
```diff
@@ -65,25 +65,156 @@ namespace kickcat
...
-                slave_error("\n update_process_data_input ERROR\n");
+                auto [obj, entry] = CoE::findObject(dict, assign_idx, i);
+                if (entry)
+                {
+                    pdo_indices.push_back(*static_cast<uint16_t*>(entry->data));
+                }
             }
         }
+
+        return pdo_indices;
```
leducp, line 119: We need to check if the mapping size if coherent with the user buffers in/out

lib/slave/src/PDO.cc
trns1997: I guess I will keep tabs of  the dictionary entry ptr in the pdoentry struct, that way we avoid this copy

lib/slave/src/PDO.cc
trns1997: I am not a fan of this method, I want to rework this to avoid having the bit_offset ref being manipulated between function calls (super C like). Need to think a better way to do this, I was thinking of encapsulating the for loop below into this function

examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
```diff
@@ -221,4 +221,4 @@ namespace kickcat::CoE
 
          return dictionary;
     }
-}
+}
```
trns1997:
```suggestion
}

```

lib/slave/src/PDO.cc
trns1997: I take it back i cannot really find an alternative. As far as the return code goes. I prefer returning a bool and then configureMapping i can return the appropriate Statuscode for INVALID_INPUT_CONFIGURATION when input mapping is problematic and INVALID_OUTPUT_CONFIGURATION for problems with output mapping

lib/slave/src/PDO.cc
trns1997: hmmm not sure if its legal?

lib/slave/include/kickcat/PDO.h
trns1997: i implemented you aliasing logic, works like a charm

lib/slave/src/PDO.cc
leducp: Pretty sure that this is legal: you can have a sensor with no RxPDO and only TxPDO for instance

examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
```diff
@@ -0,0 +1,270 @@
...
+#include "kickcat/Prints.h"
+#include "kickcat/helpers.h"
+
+using namespace kickcat;
+
+// PDO mapping configuration
+namespace pdo
+{
+    // TxPDO mapping (slave -> master): Accelerometer only
+    constexpr uint32_t tx_mapping[] = {
```
leducp: line feed before {

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -51,8 +51,8 @@ int main(int argc, char *argv[])
     mbx.enableCoE(std::move(dictionary));
 
     slave.setMailbox(&mbx);
-    pdo.setInput(buffer_in);
-    pdo.setOutput(buffer_out);
+    pdo.setInput(buffer_in, pdo_size);
```
leducp: PDO_MAX_SIZE ?

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,130 @@ int main(int argc, char *argv[])
...
+                led_r = nullptr;
+                led_g = nullptr;
+                led_b = nullptr;
+            }
+            break;
+        }
+
+        case State::SAFE_OP:
+        {
+            if (!pdo_configured)
```
leducp: not

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,130 @@ int main(int argc, char *argv[])
...
+            {
+                auto &dict = mbx.getDictionary();
+
+                StatusCode rc = pdo.configureMapping(dict);
```
leducp: I think this should be automatically done by the slave stack whenever trying to go from preop to safeop

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,130 @@ int main(int argc, char *argv[])
...
+
+                if (rc != StatusCode::NO_ERROR)
+                {
+                    printf("PDO mapping failed with code 0x%04x\n", rc);
+                    break;
+                }
+                
+                pdo_configured = true;
+
+                auto bind = [&](uint16_t idx, auto *&ptr)
```
leducp: template function to move into the slave useful helpers?

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,130 @@ int main(int argc, char *argv[])
...
-            buffer_in[4] = sensor_data.accel.z & 0xFF;
-            buffer_in[5] = (sensor_data.accel.z >> 8) & 0xFF;
-            buffer_in[6] = sensor_data.magn.x & 0xFF;
-            buffer_in[7] = (sensor_data.magn.x >> 8) & 0xFF;
-            buffer_in[8] = sensor_data.magn.y & 0xFF;
-            buffer_in[9] = (sensor_data.magn.y >> 8) & 0xFF;
-            buffer_in[10] = sensor_data.magn.z & 0xFF;
-            buffer_in[11] = (sensor_data.magn.z >> 8) & 0xFF;
+            // Cyclic PDO I/O once configured
+            if (pdo_configured)
```
leducp: if the mapping is done by the slave stack, an error shall lead to refusing moving to safeop so this check should be useless

lib/include/kickcat/CoE/OD.h
```diff
@@ -196,6 +196,7 @@ namespace kickcat::CoE
         std::string  description;
 
         void* data{nullptr};
+        bool owns_data{true};
```
leducp: I would rename it is_mapped and reverse the logic. I feel it is more readable like this 🤔

lib/slave/src/PDO.cc
trns1997: ah ofcourse

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
trns1997: hmm i just set the size based on the definition above:
```
constexpr uint32_t pdo_size = 16;

    uint8_t buffer_in[pdo_size];
    uint8_t buffer_out[pdo_size];
```

This is necessary to be able too check coherency between requested mapping and user buffer

lib/slave/src/ESMStates.cc
trns1997, line 109: I moved the PDO configuration and mapping logic into the `PreOP::routineInternal` method. Now, when the master requests a transition to `SafeOP`, the library automatically:
   - Configures by calling pdo_.configure().
   - Performs PDO remapping based on the CoE dictionary via `pdo_.configureMapping()`.
   - Validates the configuration.
   - Returns the appropriate AL Status Code if anything fails :)

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,81 @@ int main(int argc, char *argv[])
...
-            if (buffer_out[0] != 0xFF)
+            auto &dict = mbx.getDictionary();
+
+            auto bind = [&](uint16_t idx, auto *&ptr)
+            {
+                auto [obj, entry] = CoE::findObject(dict, idx, 0);
+                ptr = static_cast<std::remove_reference_t<decltype(*ptr)> *>(entry->data);
+            };
+
+            bind(0x6000, ax);
```
trns1997: now user only focuses on binding and the rest of the configuration logic has been moved to the hooks

examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
```diff
@@ -0,0 +1,272 @@
...
+        0x70000008, // LED_R, 8 bits
+        0x70010008, // LED_G, 8 bits
+        0x70020008  // LED_B, 8 bits
+    };
+    constexpr uint32_t rx_mapping_count = sizeof(rx_mapping) / sizeof(rx_mapping[0]);
+}
+
+namespace freedom
+{
+    // POC 1: Only accelerometer data
```
leducp: POC 1 ?

examples/master/freedom-k64f/freedom_k64f_mag_map_example.cc
```diff
@@ -0,0 +1,273 @@
...
+        0x70000008, // LED_R, 8 bits
+        0x70010008, // LED_G, 8 bits
+        0x70020008  // LED_B, 8 bits
+    };
+    constexpr uint32_t rx_mapping_count = sizeof(rx_mapping) / sizeof(rx_mapping[0]);
+}
+
+namespace freedom
+{
+    // POC 2: Only magnetometer data
```
leducp: POC 2? Like 'example 2'?

examples/slave/nuttx/lan9252/arduino-due/main.cc
```diff
@@ -37,8 +37,8 @@ int main(int argc, char *argv[])
         buffer_out[i] = 0xFF;
     }
 
-    pdo.setInput(buffer_in);
-    pdo.setOutput(buffer_out);
+    pdo.setInput(buffer_in, pdo_size);
```
leducp: PDO_MAX_SIZE

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
leducp: Well yes but this is a max size: the master can request less inputs/ouputs and theses buffers are the space that the slave app allocated to let the master play with the mapping.

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
leducp: Nice. Should we move the bind() helper to the lib? To avoid rewrite it again and again?

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
```diff
@@ -89,46 +89,81 @@ int main(int argc, char *argv[])
...
+            }
 
-        int ret = ioctl(led_fd, ULEDIOC_SETALL, led_set);
-        if (ret < 0)
-        {
-            int errcode = errno;
-            printf("ERROR: ioctl(ULEDIOC_SETALL) failed: %d\n", errcode);
+            if (ioctl(led_fd, ULEDIOC_SETALL, led_set) < 0)
+            {
+                printf("ERROR: ioctl(ULEDIOC_SETALL) failed: %d\n", errno);
```
leducp: You can also use strerror(errno) to have a more user friendly output (minor since this is an example)

examples/slave/nuttx/xmc4800/main_relax.cc
```diff
@@ -42,8 +42,8 @@ int main(int, char*[])
     mbx.enableCoE(std::move(dictionary));
 
     slave.setMailbox(&mbx);
-    pdo.setInput(buffer_in);
-    pdo.setOutput(buffer_out);
+    pdo.setInput(buffer_in, pdo_size);
```
leducp: PDO_MAX_SIZE (now the pdo_size constant have more impact since the master can ask for another mapping in the boundaries of theses buffers)

lib/slave/src/PDO.cc
```diff
@@ -53,37 +53,160 @@ namespace kickcat
...
+                return false;
+            }
+
+            uint32_t mapping = *static_cast<uint32_t*>(entry->data);
+
+            uint16_t index = (mapping >> 16) & 0xFFFF;
+            uint8_t  sub   = (mapping >> 8)  & 0xFF;
+            uint8_t  bits  =  mapping        & 0xFF;
+
+            if (max_size > 0 && ((bit_offset + bits + 7) / 8) > max_size)
```
leducp: and

lib/slave/src/ESMStates.cc
```diff
@@ -110,7 +110,16 @@ namespace kickcat::ESM
         {
             if (pdo_.configure() != 0)
```
leducp, line 111: pdo_.configure() shall be called after configureMapping() no? It read back the SM conf to setup internal context but if the mapping is invalid it is useless (not a major problem though since the mapping will fail anyway)

lib/slave/src/PDO.cc
```diff
@@ -53,37 +53,160 @@ namespace kickcat
...
+        {
+            auto [obj, entry] = CoE::findObject(dict, pdo_idx, i);
+            if (not entry)
+            {
+                return false;
+            }
+
+            uint32_t mapping = *static_cast<uint32_t*>(entry->data);
+
+            uint16_t index = (mapping >> 16) & 0xFFFF;
```
leducp: Maybe name constants in CoE/protocol.h to define 16, 8 and 0 and the masks

lib/src/CoE/OD.cc
```diff
@@ -246,7 +246,7 @@ namespace kickcat::CoE
 
     Entry::~Entry()
     {
-        if (data != nullptr)
+        if (data != nullptr && not is_mapped)
```
leducp: and

examples/slave/nuttx/lan9252/freedom-k64f/main.cc
trns1997: moved to slave

lib/slave/src/ESMStates.cc
trns1997: hmmm both are technically independent, so it does not matter. But i feel like its logical this way configure queries the ESC to get SM configuration. So if the SM not configured correctly no point mapping so it throws directly. So its logical to have `PDO::configure` read the ESC, and then have `PDO::configureMapping` reads the dict and remaps the buffer.

leducp (approved):

PR #268, opened 2026-01-22
Add Timer for better RT behavior

leducp:
Various minor fixes

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -22,7 +22,7 @@ class KickCATRecipe(ConanFile):
     default_options = {"shared": False, "fPIC": True, "with_esi_parser": False}
 
     # Sources are located in the same place as this recipe, copy them to the recipe
-    exports_sources = "CMakeLists.txt", "lib/*", "examples/*", "cmake/*"
+    exports_sources = "CMakeLists.txt", "lib/*", "examples/*", "cmake/*", "py_bindings/*"
 
     # def source(self):
     #     get(self, **self.conan_data["sources"][self.version], strip_root=True)
@@ -48,7 +48,7 @@ class KickCATRecipe(ConanFile):
 
         if self.settings.compiler == 'gcc' and Version(self.settings.compiler.version) < "7":
             raise ConanInvalidConfiguration("Building requires GCC >= 7")
-        
+
     def requirements(self):
         if self.options.with_esi_parser:
             self.requires("tinyxml2/10.0.0")
--- a/lib/src/OS/Unix/Timer.cc
+++ b/lib/src/OS/Unix/Timer.cc
@@ -1,0 +1,92 @@
+#include <cstdio>
+#include <stdexcept>
+#include <unistd.h>
+#include "kickcat/Error.h"
+#include "kickcat/OS/Timer.h"
+
+namespace kickcat
+{
+    Timer::Timer(nanoseconds period)
+        : period_{period}
+    {
+    }
+
+    nanoseconds Timer::period() const
+    {
+        return period_;
+    }
+
+    void Timer::start(nanoseconds sync_point)
+    {
+        // Compute the next_deadline according to the sync point (we want to tick such that one of the tick matches the sync point)
+        next_deadline_ = sync_point;
+
+        nanoseconds delta = (since_epoch() - next_deadline_);
+        if (delta < 0ns)
+        {
+            THROW_ERROR("Invalid sync point");
+        }
+
+        int64_t missing = (delta / period_);
+        next_deadline_ += missing * period_ + period_;
+
+        {
+            LockGuard lock(mutex_);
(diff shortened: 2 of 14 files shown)


conanfile.py
```diff
@@ -22,7 +22,7 @@ class KickCATRecipe(ConanFile):
     default_options = {"shared": False, "fPIC": True, "with_esi_parser": False}
 
     # Sources are located in the same place as this recipe, copy them to the recipe
-    exports_sources = "CMakeLists.txt", "lib/*", "examples/*", "cmake/*"
+    exports_sources = "CMakeLists.txt", "lib/*", "examples/*", "cmake/*", "py_bindings/*"
```
trns1997, line 25: do we really want to export this for a C++ package?

lib/src/OS/Unix/Timer.cc
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    nanoseconds Timer::period() const
+    {
+        return period_;
+    }
+
+    void Timer::start(nanoseconds sync_point)
+    {
+        // Compute the next_deadline according to the sync point (we want to tick such as one of the tick match with the sync point)
```
trns1997:
```suggestion
        // Compute the next_deadline according to the sync point (we want to tick such that one of the tick matches the sync point)
```

lib/src/OS/Unix/Timer.cc
```diff
@@ -0,0 +1,87 @@
...
+        next_deadline_ = sync_point;
+
+        nanoseconds delta = (since_epoch() - next_deadline_);
```
trns1997, line 24: just curious is, do we want to handle the case where `delta < 0`? I this case the integer devision would always truncate to 0, maybe use `floor` if we want to handle this case? But again not sure we have this use case

lib/src/OS/Unix/Timer.cc
```diff
@@ -0,0 +1,87 @@
...
+        int64_t missing = (delta / period_);
+        next_deadline_ += missing * period_ + period_;
+
+        {
```
trns1997, line 33: any reason why we scope this?

lib/src/OS/Unix/Timer.cc
```diff
@@ -0,0 +1,87 @@
...
+    {
+        period_ = period;
+    }
+
+    std::error_code Timer::wait_next_tick()
+    {
+        {
+            LockGuard lock(mutex_);
+            stop_.wait(mutex_, [&]()
+                       { return not is_stopped_; });
```
trns1997, line 64: Hmm this condition seems inverted? When the timer is running this returns `true`, causing the wait to return immediately without blocking. Not sure what the intent is here, do you want to check if the timer is stopped and return early, or wait until signaled to proceed.

Also i was wondering, maybe this method can be simplified using `wait_until`, so we can wait until the deadline or if the timer is stopped

lib/src/protocol.cc
```diff
@@ -4,6 +4,18 @@
 
 namespace kickcat
 {
+    constexpr nanoseconds THIRTY_YEARS = 30 * 365 * 24h;
```
trns1997, line 7: just curious to why choose 30 years, is it a choice, or its recommended?

conanfile.py
leducp: No but for now it blocks the conan package build and I don't want to pollute the main CMakeLists

lib/src/OS/Unix/Timer.cc
leducp: Mmm I can add a safety check: if the user ask a sync point in the past it makes no sense.

lib/src/OS/Unix/Timer.cc
leducp: Because the boolean may be shared with another thread so it need a lock guard

conanfile.py
trns1997: ahhh true that, yes in that case KISS

lib/src/protocol.cc
leducp: Standard: the EtherCAT epoch is 01/01/2000

lib/src/OS/Unix/Timer.cc
trns1997: i agree that it does not make sense, but i was just wondering cause either way this creates like a grid like tick soo techincally i could pass a past sync point and still have correct periodic grids.

lib/src/OS/Unix/Timer.cc
leducp: The idea is: if timer is stopped, it will wait until it is started somewhere else.

trns1997: ahhh thats why start() sets `is_stopped_ = false`, ok makes sense

PR #271, opened 2026-01-27
fix debug prints and refactor lan9252 interface

trns1997:
Replace lib level `printf` 's with appropriate log output. And slight refactor of the lan interface

### DIFF
--- a/lib/slave/src/ESC/Lan9252.cc
+++ b/lib/slave/src/ESC/Lan9252.cc
@@ -16,18 +16,18 @@ namespace kickcat
         spi_interface_->disableChipSelect();
         writeInternalRegister(RESET_CTL, DIGITAL_RST);
         
-        const uint16_t timeout = 10000;
+        uint16_t const number_of_retries = 10000;
         
         // Wait for reset to complete
         uint32_t reset_ctl_value;
-        if (!pollRegister(RESET_CTL, reset_ctl_value, [](uint32_t val) { return !(val & 0x1); }, timeout))
+        if (not pollRegister(RESET_CTL, reset_ctl_value, [](uint32_t val) { return not (val & 0x1); }, number_of_retries))
         {
             return -ETIMEDOUT;
         }
         
         // Check SPI interface is ready
         uint32_t byte_test_result;
-        if (!pollRegister(BYTE_TEST, byte_test_result, [](uint32_t val) { return val == BYTE_TEST_DEFAULT; }, timeout))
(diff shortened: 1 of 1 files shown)


lib/slave/src/ESC/Lan9252.cc
```diff
@@ -14,56 +14,49 @@ namespace kickcat
...
-
-        while (counter < timeout and (reset_ctl_value & 0x1))
-        {
-            counter++;
-            readInternalRegister(RESET_CTL, reset_ctl_value);
-        }
-
-        if (counter == timeout)
+        
+        const uint16_t timeout = 10000;
```
leducp: uint16_t const

lib/slave/src/ESC/Lan9252.cc
```diff
@@ -14,56 +14,49 @@ namespace kickcat
...
+        
+        // Wait for reset to complete
+        uint32_t reset_ctl_value;
+        if (!pollRegister(RESET_CTL, reset_ctl_value, [](uint32_t val) { return !(val & 0x1); }, timeout))
```
leducp: not

lib/slave/src/ESC/Lan9252.cc
```diff
@@ -14,56 +14,49 @@ namespace kickcat
...
+            return -ETIMEDOUT;
         }
-
-        if (counter == timeout)
+        
+        slave_info("Byte test read: %" PRIu32 " \n", byte_test_result);
+        
+        // Wait for device ready
+        uint32_t hw_cfg_ready;
+        if (!pollRegister(HW_CFG, hw_cfg_ready, [](uint32_t val) { return val & DEVICE_READY; }, timeout))
```
leducp: not

lib/slave/src/ESC/Lan9252.cc
```diff
@@ -14,56 +14,49 @@ namespace kickcat
...
+            slave_error("Timeout hw cfg ready \n");
+            return -ETIMEDOUT;
         }
+        
+        return 0;
+    }
 
-        if (counter == timeout)
+    template<typename Predicate>
+    bool Lan9252::pollRegister(uint16_t reg, uint32_t& value, Predicate condition, uint16_t timeout_ms)
```
leducp: Better to have timeout_ms a std::chrono::duration no?
Not sure how it is related to ms though since it is used in the for loop

trns1997: I think we can change this to use duration and timer, but yea its just to create a wait cycle, giving enough time for the hardware to take into consideration the changes requested.

PR #273, opened 2026-02-04
soem working with kickcat slaves but with a subtlity

trns1997:
### Build slaves
1. XMC
```
./scripts/build_slave_bin.sh xmc4800-relax ~/nuttxspace/nuttx && ./examples/slave/nuttx/xmc4800/deploy.sh build_xmc4800-relax/xmc4800_relax.bin
```
2. Freedom
```
./scripts/build_slave_bin.sh freedom-k64f ~/nuttxspace/nuttx && ./examples/slave/nuttx/lan9252/freedom-k64f/deploy.sh build_freedom-k64f/easycat_frdm_k64f.bi
```

### Setup
Connect Master > Freedom > XMC

###  SOEM ec_sample:
Run from soem build folder
```
sudo ./samples/ec_sample/ec_sample enp8s0 1000
```

### KickCAT easycat example:
```
sudo ./examples/master/easycat/easycat_example enp8s0
```

### DIFF
--- a/examples/master/easycat/easycat_example.cc
+++ b/examples/master/easycat/easycat_example.cc
@@ -1,6 +1,5 @@
 #include <iostream>
 #include <fstream>
-#include <iomanip>
 
 #include "kickcat/Link.h"
 #include "kickcat/Bus.h"
@@ -143,7 +142,7 @@ int main(int argc, char* argv[])
     }
 
     auto callback_error = [](DatagramState const&){ THROW_ERROR("something bad happened"); };
-    link->setTimeout(10ms);
+    link->setTimeout(10ms); // Adapt to your use case (RT loop)
 
     printf("Running loop...\n");
 


lib/slave/src/ESMStates.cc
```diff
@@ -198,10 +198,10 @@ namespace kickcat::ESM
         pdo_.updateInput();
         pdo_.updateOutput();
 
-        if (currentStatus.hasExpiredWatchdog())
-        {
-            return Context::build(State::SAFE_OP, SYNC_MANAGER_WATCHDOG);
-        }
+        // if (currentStatus.hasExpiredWatchdog())
```
trns1997: kickcat slaves expect new data = reset watchdog and the kickcat master resets the timer so everything works well. Where are SOEM does not reset the timer so it keep counting up until it expires. Commenting this out ignores the hardware watchdog timer and forces the slave to stay in OP. I am not sure what needs to be done exactly so that kickcat slaves natively work with SOEMs examples?

leducp: This part of the slave was tested with the official conformance test tool so I doubt that this is the solution. Maybe the example from SOEM need to be tweaked?

leducp: BTW the timer is reset through a logical access, it is not a direct watchdog pumping so something fishy here

trns1997: Hmmm https://github.com/leducp/KickCAT/blob/479e89d05b732073fdcd255e93c563798c9d2fb7/lib/master/src/Bus.cc#L662 this might be the reason not sure, it seems to explain what we're are seeing and in the forum they seem to have something similar https://support.hms-networks.com/hc/fr/articles/8749732317586-Comment-rem%C3%A9dier-%C3%A0-l-erreur-de-temps-mort-Sync-Manager-Watchdog-lors-de-l-utilisation-de-TwinCAT-dans-un-r%C3%A9seau-EtherCAT

But you are right it passed the official conformity test so it's surprising that SOEM doesn't work by default 🤔

trns1997: Alright so our slave eeprom are not defined correctly, and SOEM does not set the watchdog trigger manually it trusts the contents of the eeprom to configure itself. Here is a minimal patch where SOEM sets the watchdog trigger that way it can run on ourself: https://github.com/trns1997/SOEM/commit/8d830a6a2a95602cb3c8a4dfe18256881fdb19ff

examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
```diff
@@ -28,11 +28,11 @@ namespace kickcat::CoE
                 "Identity Object",
                 {}
             };
-            CoE::addEntry<uint8_t>(object,0,8,0,7,static_cast<CoE::DataType>(5),"Subindex 000",0x4);
```
trns1997, line 31: @leducp i know the file is autogenerated, but i find the version on the right more readable. Plus i had done it manually, so i suggest the newer version for readability sake, what do you think?

leducp: od_generator is a helper, if a manual version is better let's keep it. However it means that we should improve od_generator to have a more readable output imho (I hate generator that produce unreadable output, and our current generator is not really good in this sense).

examples/master/easycat/easycat_example.cc
```diff
@@ -137,25 +143,16 @@ int main(int argc, char* argv[])
     }
 
     auto callback_error = [](DatagramState const&){ THROW_ERROR("something bad happened"); };
-    link->setTimeout(10ms);  // Adapt to your use case (RT loop)
```
leducp, line 140: Keep comment

examples/master/easycat/easycat_example.cc
```diff
@@ -164,51 +161,30 @@ int main(int argc, char* argv[])
...
-                printf("\n -*-*-*-*- slave %u -*-*-*-*-\n %s", easycat.address, toString(easycat.error_counters).c_str());
-            }
-
-            microseconds sample = duration_cast<microseconds>(t4 - t3 + t2 - t1);
-            std::string sample_str = std::to_string(sample.count());
-            fwrite(sample_str.data(), 1, sample_str.size(), stat_file);
-            fwrite("\n", 1, 1, stat_file);
+            // Move cursor up
+            printf("\033[%zuA", bus.slaves().size());
+            fflush(stdout);
```
leducp, line 180: Be cautious: flushing here impact your jitter

examples/master/easycat/easycat_example.cc
```diff
@@ -1,12 +1,12 @@
 #include <iostream>
 #include <fstream>
+#include <iomanip>
```
leducp: Why do you need that :)

trns1997: Ugggh it's my IDE it keeps adding iomanip when i format the file :/

examples/slave/nuttx/lan9252/freedom-k64f/od_populator.cc
trns1997: Created issue #276

examples/master/easycat/easycat_example.cc
trns1997: Yes i know but UX wise its nicer. In cases where we want to benchmark, we should avoid using this example. I will consider creating a benchmarking script that minimizes prints and other such operation that needlessly impact performance metrics

leducp (approved):

PR #275, opened 2026-02-07
small bug fix, luckily no visible impact yet

trns1997:


lib/slave/src/ESC/Lan9252.cc
```diff
@@ -204,7 +204,7 @@ namespace kickcat
         // CSR_DATA is 4 bytes
         uint32_t padding = 0;
         std::memcpy(&padding, data, size);
-        writeInternalRegister(ECAT_CSR_DATA, data, sizeof(padding));
+        writeInternalRegister(ECAT_CSR_DATA, &padding, sizeof(padding));
```
leducp, line 207: Nice catch. Did you had the time to check on real hardware?

leducp (approved):

lib/slave/src/ESC/Lan9252.cc
trns1997: Yup tested and approved

Issue #218, opened 2025-09-25
Quality


leducp:
Generic issue to link to PR that improve code quality:
* code coverage
* integration test (if no better naming)
* CI (if no better naming)
* documentation (typos, phrasing, NOT new sections)

PR #288, opened 2026-02-24
Cleanup

leducp:
related to #218

### DIFF
--- a/lib/slave/src/ESC/EmulatedESC.cc
+++ b/lib/slave/src/ESC/EmulatedESC.cc
@@ -386,7 +386,7 @@ namespace kickcat
         {
             case eeprom::Control::READ:
             {
-                if (memory_.eeprom_address < eeprom_.size())
+                if (memory_.eeprom_address + 2 <= eeprom_.size())
                 {
                     std::memcpy((void*)&memory_.eeprom_data, eeprom_.data() + memory_.eeprom_address, 4);
                 }
@@ -594,14 +594,14 @@ namespace kickcat
             return; // watchdog deactivated
         }
 
-        if ((lastLogicalWrite_ + delay) < since_epoch())
+        if ((lastLogicalWrite_ + delay) >= since_epoch())
         {
-            // Watchog OK
+            // Watchdog OK: last write is recent enough
             memory_.watchdog_status_process_data = 0;
             return;
         }
 
-        // Watchog Triggered - detect front
+        // Watchdog triggered - detect front
         if (memory_.watchdog_status_process_data == 0)
         {
             // Increase PDO wdg counter and reflect the current state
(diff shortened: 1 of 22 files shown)


lib/slave/src/ESC/EmulatedESC.cc
```diff
@@ -386,7 +386,7 @@ namespace kickcat
         {
             case eeprom::Control::READ:
             {
-                if (memory_.eeprom_address < eeprom_.size())
+                if (memory_.eeprom_address + 2 <= eeprom_.size())
```
trns1997, line 389: Is there an explaination for this ?

lib/slave/src/ESC/EmulatedESC.cc
```diff
@@ -594,14 +594,14 @@ namespace kickcat
             return; // watchdog deactivated
         }
 
-        if ((lastLogicalWrite_ + delay) < since_epoch())
+        if ((lastLogicalWrite_ + delay) >= since_epoch())
```
trns1997, line 597: this is a logic flip, is this correct? The impact needs to be tested?

lib/src/Frame.cc
```diff
@@ -47,12 +47,15 @@ namespace kickcat
...
     Frame& Frame::operator=(Frame&& other)
     {
+        auto next_offset = other.next_datagram_ - other.first_datagram_;
+        auto last_offset = other.last_datagram_ - other.first_datagram_;
+
         frame_ = std::move(other.frame_);
         ethernet_ = pointData<EthernetHeader>(frame_.data());
         header_   = pointData<EthercatHeader>(ethernet_);
         first_datagram_ = pointData<uint8_t>(header_);
-        next_datagram_  = other.next_datagram_;
```
trns1997, line 54: oh wow this was a big bug, other was empty after the move right?

lib/slave/src/ESC/EmulatedESC.cc
leducp: Well it is to avoid an out of bound access: the current check does not look for the data that are going to be read.

lib/slave/src/ESC/EmulatedESC.cc
leducp: The current seems to be wrong: last access + watchdog delay >= current time means no watchdog trigger (as the current comment). However I'm surprised that it ever works (or we never test the watchdog with the emulation but I'm pretty sure we do) so I vote for a test

lib/src/Frame.cc
leducp: yeah... the move was wrong and the unit tests were working on dangling pointers...

lib/slave/src/ESC/EmulatedESC.cc
trns1997: its only used in network simulation soo, impact on hardware none :)

PR #295, opened 2026-02-26
Master testing with network sim

### LINKED ISSUE #293: [CI] Add network simulator stage

trns1997:
Closes #293

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,14 +148,14 @@ jobs:
 
           echo "Test passed!"
 
-      # - name: Upload Simulation Logs
-      #   uses: actions/upload-artifact@v4
-      #   if: failure()
-      #   with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -71,13 +71,92 @@ jobs:
...
+          if grep -q "input_slave_0:" test_output.log; then
+               echo "SUCCESS: ${{ matrix.test_case.name }} received input data."
+          else
+              echo "FAILURE: ${{ matrix.test_case.name }} did NOT receive input data."
+              exit 1
+          fi
+
+          echo "Test passed!"
+
+      # - name: Upload Simulation Logs
```
trns1997: I am hesitant on the fact that we want these logs. What do you think @leducp

leducp: Better to have them for now: we can disable this upload later if it does not help.

Issue #126, opened 2023-10-13
Distributed clocks / SYNC message


thetooth:
I'm currently working on a robotics project and have been entertaining the idea of replacing CODESYS which I have been using for the RT aspects of the system with an open source stack. Found this library and it's very cool, took maybe all of an hour to adapt the example and get my drive booted and PDO mapped. The problem I now face is its a Delta B3 drive and apparently wont move unless it's receiving some form of sync message.

`Delta servo drives only support synchronous operation which is the controller periodically sends the SYNC object (COB-ID = 0x80).`

The drive is fully powered up (locked shaft), receiving control word, target position and mode of operation, it's sending back all the right status word bits as well as actual position + velocity. But it refuses to budge. The only configuration difference between this and CODESYS I can see is that its not receiving a distributed clock.

Now the problem is that I only know a little bit about CoE, and next to nothing about EtherCAT itself. Reading what's out there it seems as though SYNC messages and distributed clocks are two different things, if that's true is there a quick and dirty way to generate that payload and get this project moving forward? I assume the only reason we don't already have DC in this library is because it's a non trivial task and unfortunately I'm probably not in the position to contribute one even though I would like to.

Any help would be much appreciated.

leducp:
DC sync is indeed not trivial, but it is not that complicated too. The main reason of why it is not implemented yet is the lack of setup to test it thoroughly which is a really important aspect for us as the stack is used in industrial and critical software.

About your problem: SYNC is a CAN concept that may be mapped to DC sync when using CoE profile (CAN over EtherCAT) but as my understanding of this part is scarce, it may also be mapped to something else.

Can you send me the manual of your drive?  Also can you do a network trace (wireshark or tcpdump) with CODESYS and with KickCAT from start to functional state (or blocked state with KickCAT)? With this I'm pretty sure that we will be able to at least workaround this situation :)

thetooth:
Thanks, you can find what you're after here: https://thetooth.name/tmp/

Ignore the FRMW commands in the kickcat dump, that's just wishful thinking on my part.

leducp:
It seems that your module documentation have an update:
https://filecenter.deltaww.com/Products/download/06/060201/Manual/DELTA_IA-ASD_ASDA-B3_UM_EN_20230428.pdf

It says that EtherCAT mode can be either free run or DC so it should work without DC enabled (12.2.2.1 )
Can you confirm that the doc is just an update and is applicable to your case?

[...]

PR #296, opened 2026-02-26
Add exprimental DC support

leducp:
related to #126

### DIFF
--- a/examples/master/marvin/marvin.cc
+++ b/examples/master/marvin/marvin.cc
@@ -15,8 +15,10 @@
 #include "CanOpenErrors.h"
 #include "MarvinProtocol.h"
 
-//#include <rtm/probe.h>
-//#include <rtm/io/posix/local_socket.h>
+constexpr double REDUCTION_RATIO[] = {120.0, 120.0, 100.0, 100.0, 100.0, 100.0, 100.0,
+                                        120.0, 120.0, 100.0, 100.0, 100.0, 100.0, 100.0};
+constexpr double ENCODER_TICKS_PER_TURN[] = {1<<20, 1<<20, 1<<19, 1<<19, 1<<19, 1<<19, 1<<19,
+                                            1<<20, 1<<20, 1<<19, 1<<19, 1<<19, 1<<19, 1<<19};
 
 using namespace kickcat;
 
@@ -28,22 +30,7 @@ int main(int argc, char *argv[])
         printf("usage no redundancy mode : ./test NIC_nominal\n");
         return 1;
     }
-/*
-    rtm::Probe probe;
-    {
-        auto io = std::make_unique<rtm::LocalSocket>();
-        auto rc = io->open(rtm::access::Mode::READ_WRITE);
-        if (rc)
-        {
-            printf("io open() error: %s\n", rc.message().c_str());
-            return 1;
-        }
 
-        probe.init("marvin", "op",
-                since_epoch(), 1ms, 42,
-                std::move(io));
-    }
-*/
     std::string red_interface_name = "";
     std::string nom_interface_name = argv[1];
     if (argc == 3)
@@ -77,7 +64,22 @@ int main(int argc, char *argv[])
     Bus bus(link);
     nanoseconds sync_point = 0ns;
 
-    uint8_t io_buffer[2048];
+    struct Motor
+    {
+        Slave* slave{};
+        pdo::Output* output{};
+        pdo::Input*  input{};
+        CoE::CiA::DS402::StateMachine state_machine{};
+        int32_t initial_position{};
+        double reduction_ratio;
+        double encoder_ticks_per_turn;
+        uint16_t prev_status_word{0xFFFF};
+        uint16_t prev_control_word{0xFFFF};
+    };
+    std::vector<Motor> motors;
+    motors.resize(14);
+
+    uint8_t io_buffer[2048] = {};
     try
     {
         bus.init(100ms);
@@ -94,26 +96,46 @@ int main(int argc, char *argv[])
         bus.requestState(State::PRE_OP);
         bus.waitForState(State::PRE_OP, 3000ms);
 
-
-        /*
-        // Map RxPDO/TxPDO
         for (int i = 0; i < 7; ++i)
         {
-            mapPDO(bus, bus.slaves().at(i), 0x1A01, pdo::rx_mapping, pdo::rx_mapping_count, 0x1C13);
-            mapPDO(bus, bus.slaves().at(i), 0x1601, pdo::tx_mapping, pdo::tx_mapping_count, 0x1C12);
+            motors[i].slave = &bus.slaves().at(i + 1);
+            motors[i].encoder_ticks_per_turn = ENCODER_TICKS_PER_TURN[i];
+            motors[i].reduction_ratio = REDUCTION_RATIO[i];
         }
-        */
-
-
-        uint8_t mode = 8;
-        uint32_t mode_size = 1;
-        for (int j = 1; j < 8; ++j)
+        for (int i = 7; i < 14; ++i)
         {
-            bus.writeSDO(bus.slaves().at(j), 0x6060, 0, Bus::Access::PARTIAL, &mode, mode_size);
+            motors[i].slave = &bus.slaves().at(i + 2); // skip first and last of first arm
+            motors[i].encoder_ticks_per_turn = ENCODER_TICKS_PER_TURN[i];
+            motors[i].reduction_ratio = REDUCTION_RATIO[i];
+        }
+
+        for (auto& motor : motors)
+        {
+            mapPDO(bus, *motor.slave, 0x1A01, pdo::tx_mapping, pdo::tx_mapping_count, 0x1C13);
+            mapPDO(bus, *motor.slave, 0x1601, pdo::rx_mapping, pdo::rx_mapping_count, 0x1C12);
         }
 
         printf("mapping\n");
         bus.createMapping(io_buffer);
+
+        for (auto& motor : motors)
+        {
+            motor.output = reinterpret_cast<pdo::Output *>(motor.slave->output.data);
+            motor.input  = reinterpret_cast<pdo::Input *> (motor.slave->input.data);
+            motor.state_machine.disable();
+
+            uint8_t mode = 8;   //CSP
+            uint32_t mode_size = 1;
+            bus.writeSDO(*motor.slave, 0x6060, 0, Bus::Access::PARTIAL, &mode, mode_size);
+
+            uint8_t interpolation_value = 1;
+            uint32_t interpolation_value_size = 1;
+            bus.writeSDO(*motor.slave, 0x60C2, 1, Bus::Access::PARTIAL, &interpolation_value, interpolation_value_size);
+
+            int8_t interpolation_index = -3; // 10^(-3) = milliseconds
+            uint32_t interpolation_index_size = 1;
+            bus.writeSDO(*motor.slave, 0x60C2, 2, Bus::Access::PARTIAL, &interpolation_index, interpolation_index_size);
+        }
 
         printf("Request SAFE OP\n");
         bus.requestState(State::SAFE_OP);
@@ -143,25 +165,6 @@ int main(int argc, char *argv[])
     {
         //printf("- previous error was a false alarm - ");
     };
-
-    constexpr int MOTORS = 14;
-    pdo::Output* output_pdo[MOTORS];
-    pdo::Input*  input_pdo[MOTORS];
-    CoE::CiA::DS402::StateMachine state_machine[MOTORS];
-    for (int i = 0; i < 7; ++i)
-    {
-        Slave& motor = bus.slaves().at(i + 1);
-        output_pdo[i] = reinterpret_cast<pdo::Output *>(motor.output.data);
-        input_pdo[i] = reinterpret_cast<pdo::Input *>(motor.input.data);
-        state_machine[i].disable();
-    }
-    for (int i = 0; i < 7; ++i)
-    {
-        Slave& motor = bus.slaves().at(i + 9);
-        output_pdo[i+7] = reinterpret_cast<pdo::Output *>(motor.output.data);
-        input_pdo[i+7] = reinterpret_cast<pdo::Input *>(motor.input.data);
-        state_machine[i+7].disable();
-    }
 
     Timer timer{1ms};
     timer.start(sync_point);
@@ -201,36 +204,28 @@ int main(int argc, char *argv[])
 
     link->setTimeout(1500us);
 
-    //printf("mapping: input(%d) output(%d)\n", motor.input.bsize, motor.output.bsize);
-
-    int32_t initial_position[MOTORS] = {0};
     nanoseconds start_time = kickcat::since_epoch();
-
-    std::shared_ptr<mailbox::request::AbstractMessage> msg = nullptr;
-
     constexpr int64_t LOOP_NUMBER = 12 * 3600 * 1000; // 12h
     //constexpr int64_t LOOP_NUMBER = 1000 * 60 * 5; // 5min
     int64_t last_error = 0;
     for (int64_t i = 0; i < LOOP_NUMBER; ++i)
     {
-        //probe.log();
-
         try
         {
             if (i == 500)
             {
-                for (int j = 0; j < MOTORS; ++j)
+                for (auto& motor : motors)
                 {
-                    output_pdo[j]->target_position = input_pdo[j]->actual_position;
-                    state_machine[j].enable();
+                    motor.output->target_position = motor.input->actual_position;
+                    motor.state_machine.enable();
                 }
             }
 
             if (i == 4000)
             {
-                for (int j = 0; j < MOTORS; ++j)
+                for (auto& motor : motors)
                 {
-                    initial_position[j] = input_pdo[j]->actual_position;
+                    motor.initial_position = motor.input->actual_position;
                 }
                 start_time = kickcat::since_epoch();
             }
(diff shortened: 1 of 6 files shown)


cmake/debug.cmake
```diff
@@ -141,3 +141,19 @@ option(DEBUG_SPI_INFO    "Enable spi debug info level traces" OFF)
...
+
+option(DEBUG_DC_WARNING "Enable distributed clock warning level traces" OFF)
+if (DEBUG_DC_WARNING)
+  add_compile_definitions(DEBUG_DC_WARNING)
+endif()
+
+option(DEBUG_DC_INFO    "Enable distributed clock info level traces" OFF)
+if (DEBUG_DC_INFO)
+  add_compile_definitions(DEBUG_DC_INFO)
+endif()
```
trns1997:
```suggestion
endif()

```

examples/master/freedom-k64f/freedom_k64f_static_map_example.cc
```diff
@@ -133,7 +136,14 @@ int main(int argc, char* argv[])
         printESC(slave);
     }
 
-    auto& easycat = bus.slaves().at(0);
+    auto& easycat = bus.slaves().at(1);
```
trns1997: revert this when time come

examples/master/ingenia/ingenia_info.cc
```diff
@@ -156,11 +156,9 @@ int main(int argc, char *argv[])
     printEntryDescription(bus, ingenia, 0x1018, 0,
         CoE::SDO::information::ValueInfo::MAXIMUM);
 
-    printEntryDescription(bus, ingenia, 0x1600, 0,
-        CoE::SDO::information::ValueInfo::DEFAULT);
-
-    printEntryDescription(bus, ingenia, 0x1600, 0,
-        CoE::SDO::information::ValueInfo::DEFAULT | CoE::SDO::information::ValueInfo::MINIMUM | CoE::SDO::information::ValueInfo::MAXIMUM);
+    printEntryDescription(bus, ingenia, 0x100A, 0, CoE::SDO::information::ValueInfo::DEFAULT);
```
trns1997: not sure why this was changed? Did you test this on an ingenia?

examples/master/marvin/marvin.cc
```diff
@@ -0,0 +1,325 @@
...
+        printf("now %f, sp: %f -> %f\n",
+            seconds_f(since_epoch()).count(),
+            seconds_f(sync_point).count(),
+            seconds_f(sync_point - since_epoch()).count());
+
+        bus.requestState(State::PRE_OP);
+        bus.waitForState(State::PRE_OP, 3000ms);
+
+
+        auto map_pdos = [&](uint8_t id)
```
trns1997: ah high time we create a global method to remap. I have created an issue #302 :)

lib/master/include/kickcat/Bus.h
```diff
@@ -81,6 +89,16 @@ namespace kickcat
         void checkMailboxes( std::function<void(DatagramState const&)> const& error);
         void processMessages(std::function<void(DatagramState const&)> const& error);
 
+        // DC
```
trns1997: doc

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+
+namespace kickcat
+{
+    nanoseconds Bus::enableDC(nanoseconds cycle_time, nanoseconds shift_cycle, nanoseconds start_delay)
+    {
+        for (auto& slave : slaves_)
+        {
+            if (slave.esc.features & ESC::feature::DC_AVAILABLE)
+            {
+                dc_slave_ = &slave;
```
trns1997, line 17: If i understand correctly, we have 1 designated slave as the DC reference (the first one we find) and all other slave sync to it right? And this assumes that slaves in the vector are already ordered by bus topology, i am guessing this is a safe assumption but i just want to be sure :). Or maybe we loop throught all slave and get the slave with lowest physical address to make sure we have the lowest delay and the best slave?

examples/master/ingenia/ingenia_info.cc
leducp: No but this is a standard CANOpen register. However not related so to revert before merge.

lib/master/src/dc.cc
leducp: Yes: the slaves are loaded in the discovery order, which is the order of frame processing (which impact the DC). Edit (because I answer only partially): the slaves that can be synchronized must be behind the DC reference slave, that's why the system is looking for the first one it encounter in the network.

lib/src/protocol.cc
```diff
@@ -4,16 +4,16 @@
 
 namespace kickcat
 {
-    constexpr nanoseconds THIRTY_YEARS = 30 * 365 * 24h;
+    constexpr nanoseconds ECAT_EPOCH_OFFSET = 10957 * 24h; // 1970-01-01 to 2000-01-01 (30 years + 7 leap days)
```
trns1997, line 7: you need to add a new offset which is for DC `Universal zero point 1.1.2000 00:00` See https://infosys.beckhoff.com/english.php?content=../content/1033/ethercatsystem/2469118347.html&id=  in **Summary of technical data**

lib/master/include/kickcat/Bus.h
```diff
@@ -81,6 +89,16 @@ namespace kickcat
...
+        void sendDriftCompensation(std::function<void(DatagramState const&)> const& error);
+
+        /// \brief  Check if distributed clocks are synchronized
+        /// \details Reads the system time difference register (0x092C) of each DC slave.
+        ///          The value converges to zero when the slave is synchronized with the reference clock.
+        /// \param  threshold  Maximum acceptable time difference in nanoseconds
+        /// \return true if all DC slaves are synchronized within the given threshold
+        bool isDCSynchronized(nanoseconds threshold = 1000ns);
```
trns1997: i this this tolerance should be equal to `100ns`, `The EtherCAT master automatically synchronizes each local clock with the reference clock in the EtherCAT system with a precision of < 100 ns, irrespective of the distance between individual EtherCAT slaves` from the **Summary of technical data**

trns1997: is this `threshold` equivalent to the `tolerance` they mentioned here? I am not super sure :\

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+        dc_slave_->delay = 0ns;
+
+        // Process all slaves in order
+        uint16_t parenthold = 0;
+        for (auto& slave : slaves_)
+        {
+            if (slave.isDCSupport())
+            {
+                // Compute time offset from ref clock
+                slave.dc_time_offset = master_time - slave.dc_ecat_received_time;
```
trns1997, line 222: doesn't this rewrite content of https://github.com/leducp/KickCAT/pull/296/changes#diff-d168d20b282617d9dd163ddee7abb33e356bfacae166e84ce92f7e04f5db578fR208 [line 208] above?

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+                                                                             entry_ports[parent_slave->address]);
+                    }
+
+                    // ESC-specific difference between processing and forwarding delay.
+                    // Should be parsed from the ESI per slave - 0 as fallback.
+                    constexpr nanoseconds tDiff = 0ns;
+
+                    // If child's internal round-trip exceeds parent's view of the branch,
+                    // the subtraction would yield a negative delay - negate dt1 to compensate.
+                    if (entry_to_prev_delta > parent_port_to_prev_delta)
```
trns1997, line 370: this is seems hackish, this should not be possible and it this happens then the bus slave topology was wrongly determined

lib/src/protocol.cc
leducp: Well, this is the offset you are talking about so I'm not sure if I'm missing something 🤔

lib/master/include/kickcat/Bus.h
leducp: This threshold is the tolerance for the client to consider that the network is synchronized with DC. In theory you can have <100ns of sync but it depends on a few parameters:
- the refresh rate of the system clock
- the refresh rate of the DC clock
- the quality of the propagation delay (the current version is not perfect : it is missing the ESC delay which can have a big impact on medium to big network size or with complex topology

Here I choose 1000ns / 10us which is a usual acceptable jitter for a RT system @ 1KHz, but one can seelect either a stricter or looser threshold.

lib/src/protocol.cc
trns1997: ah sorryyyyy you are right i misunderstood the comment. Its unix to ecat time. My bad this is good

lib/master/src/dc.cc
leducp: Yes, but I felt that it was clearer to setup the DC reference when I wrote it.

lib/master/src/dc.cc
leducp: Well, no : you can catch a register clock that overflow during the frame propagation (the one that trigger the record time), leading to negative time.

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+
+        //----------------------- Apply cycle time ----------------------//
+        uint32_t cycle_time_raw = static_cast<uint32_t>(cycle_time.count());
+        broadcastWrite(reg::DC_SYNC0_CYCLE_TIME, &cycle_time_raw, 4);
+
+        //------------------------ Set start time -----------------------//
+        // 1. Get current network time to compute start time (relative to absolute)
+        uint64_t network_time;
```
trns1997:
```suggestion
        uint64_t network_time = 0;
```

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+        };
+
+        for (auto& slave : slaves_)
+        {
+            if (not slave.isDCSupport())
+            {
+                continue;
+            }
+            int64_t raw_dc_time_diff = slave.dc_time_offset.count();
+            if (raw_dc_time_diff == 0)
```
trns1997: i get this is impossible but why is this invalid?

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+                }
+
+                // Only calculate propagation delay if slave has a DC parent and is not the DC reference
+                if (parent_slave != nullptr and &slave != dc_slave_)
+                {
+                    // Find port on parent this slave is connected to
+                    uint8_t parent_port = parentPort(consumed_ports[parent_slave->address]);
+
+                    // If parent has only one link (topology == 1), use parent's entry port
+                    if (parent_slave->countOpenPorts() == 1)
```
trns1997, line 310: I was just reading through the definition of SOEM for DC for comparison sakes, and found that we do not have the same definition for determining the topology. Here is the equivalent section from what i understand in SOEM: https://github.com/OpenEtherCATsociety/SOEM/blob/b410bf6ef599d5c85302ea45cae5f55f8e9aa394/src/ec_dc.c#L366. We in kickcat only check the physical link to determine the topology: https://github.com/leducp/KickCAT/blob/e8e462eec2a55cc8597340aa0bd6fb68d7a4eb6a/lib/master/src/Slave.cc#L59 where as SOEM checks if the port is open and the com is established which seems to be correct to me. What do you think?

lib/master/src/dc.cc
```diff
@@ -0,0 +1,554 @@
...
+                        {
+                            parent_index = i;
+                        }
+                    }
+
+                    if (current_index > parent_index + 1)
+                    {
+                        for (size_t i = parent_index + 1; i < current_index; ++i)
+                        {
+                            if (slaves_[i].isDCSupport() and
```
trns1997, line 349: Note SOEM checks for any slave between parent and child, not just DC capable ones. See https://github.com/OpenEtherCATsociety/SOEM/blob/b410bf6ef599d5c85302ea45cae5f55f8e9aa394/src/ec_dc.c#L360 and https://github.com/OpenEtherCATsociety/SOEM/blob/b410bf6ef599d5c85302ea45cae5f55f8e9aa394/src/ec_dc.c#L391

where as we do not do this, we do not consider topologies where we have 2 non dc slave between 2 DC enabled slaves.

lib/master/src/dc.cc
leducp: Maybe: I'm not sure how an open port cannot have an established link though 🤔 
So far I didn't see a difference between the two feedback: Im probably missing something but I wonder what

lib/master/src/dc.cc
leducp: Agree but why should we care for non DC slaves? The delay is computed by using the timestamp of the DC slaves so we can skip the ones in the middles no?

lib/master/src/dc.cc
leducp: It means that the slave do not generate its own clock: if I understand correctly, some old DC slaves are not full features. However maybe this is not needed at all: need to recheck on the robot I have because I may have misunderstood a behavior.

lib/master/src/dc.cc
trns1997: Even though non DC slaves don't have time stamp they introduce propagation delay which needs to be considered so that the DC algorithm compensates the right amount. And SOEM seems to keep this fact into consideration.

leducp: Bus the propagation delay is compute through the timestamps recorded by the slaves : you can skip whatever is in the middle by substracting the timestamps of two DC slaves.

lib/master/src/dc.cc
trns1997: Note section **4.1Link status** of Section 1 of the doc states:
```
The link status of each port is available in the ESC DL status register (0x0110:0x0111), most
important are the “communication established” bits 15, 13, 11, and 9.
```

Further more in section **5.6.1.3 MI link detection and configuration** it says that physical link can be established but the com may not be active (just a scenarios where physical link is up but communication is NOT established):
```
Since the MI link detection does not solely rely on the PHY link status bit (register 1[2]), the local PHY
and the remote PHY may indicate a link, but the ESC refuses it because it does not fulfill EtherCAT
requirements.
```

I think to determine open port or more important the topology we need to consider both ESC DL status register (0x0110[15,13,11,9] – Communication established), and ESC DL status register (0x0110[14,12,10,8] – loop open/closed).

lib/master/src/dc.cc
trns1997: So what they call `wire delay` is taken care of? Here is says:
```
Slaves without distributed clocks support:
Slaves with max. 2 ports do not have to support DC features. Processing/forwarding delay of such
slaves is treated like a “wire delay” by the surrounding DC capable slaves.
```
Just the max 2 ports thing does not make sense to me. If a slave has more than 2 ports it most definately supports DC features?

lib/master/src/dc.cc
leducp: OK : I'll open an issue to fix this but I'll keep this modification outside this PR as it is not directly related.

lib/master/src/dc.cc
leducp: Yes, because it can be a Y or a X in the topology, which means it is a root of the delay computation for the branches it connects

lib/master/src/dc.cc
leducp: Seems to ave no effect (as expected), removed (probably a remains of my debug time)

examples/master/marvin/marvin.cc
```diff
@@ -0,0 +1,304 @@
...
+            seconds_f(sync_point).count(),
+            seconds_f(sync_point - since_epoch()).count());
+
+        bus.requestState(State::PRE_OP);
+        bus.waitForState(State::PRE_OP, 3000ms);
+
+
+        /*
+        // Map RxPDO/TxPDO
+        for (int i = 0; i < 7; ++i)
```
trns1997, line 99: you want to keep these?

examples/master/marvin/marvin.cc
```diff
@@ -0,0 +1,304 @@
...
+int main(int argc, char *argv[])
+{
+    if (argc != 3 and argc != 2)
+    {
+        printf("usage redundancy mode : ./test NIC_nominal NIC_redundancy\n");
+        printf("usage no redundancy mode : ./test NIC_nominal\n");
+        return 1;
+    }
+/*
+    rtm::Probe probe;
```
trns1997: debug probes you want to keep?

lib/src/OS/Unix/Timer.cc
```diff
@@ -84,8 +82,14 @@ namespace kickcat
         next_deadline_ += period_;
         if (next_deadline_ < last_wakeup_)
         {
-            return std::error_code{ETIME, std::system_category()};
+            printf("!!! LATE !!!\n");
```
trns1997: revert

lib/src/OS/Unix/Timer.cc
```diff
@@ -84,8 +82,14 @@ namespace kickcat
...
         }
+        //while (next_deadline_ < last_wakeup_)
```
trns1997: remove?

lib/src/OS/Windows/Timer.cc
```diff
@@ -0,0 +1,81 @@
...
+        if (next_deadline_ > now)
+        {
+            std::this_thread::sleep_for(next_deadline_ - now);
+        }
+        last_wakeup_ = since_epoch();
+
+        next_deadline_ += period_;
+        if (next_deadline_ < last_wakeup_)
+        {
+            printf("!!! LATE !!!\n");
```
trns1997, line 76: throw or better error message?

unit/src/bus-t.cc
```diff
@@ -43,8 +43,8 @@ class BusTest : public testing::Test
         // detectSlaves: broadcastRead
         mock_link->handleWriteThenRead(Command::BRD, 1);
 
-        // resetSlaves: 10 broadcastWrites + 3 broadcastReads
-        for (int i = 0; i < 10; ++i)
+        // reset slaves
+        for (int i = 0; i < 11; ++i)
```
trns1997, line 47: why this change?

examples/master/marvin/marvin.cc
leducp: Yes: this example will evolve again later

examples/master/marvin/marvin.cc
leducp: Same: we need an official real time monitor export before re-enable it
Note: I want to cleanup before the release

lib/src/OS/Unix/Timer.cc
leducp: No :)
I don't know how to handle the problem right now because the timer is always late at the first tick in a DC setup

lib/src/OS/Unix/Timer.cc
leducp: good catch

lib/src/OS/Windows/Timer.cc
leducp: no throw as this is a common problem, not an exception, but dunno if an error code is the best way to handle it right now

unit/src/bus-t.cc
leducp: I added stuff in the init remember :) (fetchDL mainly)

trns1997: ah right

lib/src/OS/Unix/Timer.cc
trns1997: don't forget to uncomment this :)

trns1997 (approved):

PR #301, opened 2026-02-27
Refactor bus unit tests

### LINKED ISSUE #300: Refactor bus unit tests
Bus unit test currently rely on mocked socket, which lead to a high complexity code and reduce the unit testing approach as bus never work directly with the socket.

Mock the link to work on it instead.

leducp:
close #300

### DIFF
--- a/lib/master/include/kickcat/AbstractLink.h
+++ b/lib/master/include/kickcat/AbstractLink.h
@@ -1,0 +1,63 @@
+#ifndef KICKCAT_ABSTRACT_LINK_H
+#define KICKCAT_ABSTRACT_LINK_H
+
+#include <cstring>
+#include <functional>
+
+#include "kickcat/Frame.h"
+
+namespace kickcat
+{
+    /// \brief Abstract interface for the EtherCAT link layer.
+    /// \details The link layer is responsible for sending and receiving EtherCAT frames on the wire.
+    ///          It provides two communication modes:
+    ///          - Synchronous: writeThenRead() sends a frame and blocks until the response is received.
+    ///          - Asynchronous: addDatagram() queues datagrams with associated callbacks, then
+    ///            processDatagrams() sends them all and dispatches the responses through the callbacks.
+    class AbstractLink
+    {
+    public:
+        virtual ~AbstractLink() = default;
+
+        /// \brief Send a frame on the wire and wait for the response (synchronous).
+        /// \param frame Frame to send - modified in place with the response data.
+        virtual void writeThenRead(Frame& frame) = 0;
+
+        /// \brief Queue a datagram to be sent on the next processDatagrams() call.
+        /// \param command   EtherCAT command (e.g. BRD, FPRD, FPWR, LRD, LWR, LRW).
+        /// \param address   Slave address (auto-increment, configured, or logical depending on command).
+        /// \param data      Pointer to the data payload to send.
+        /// \param data_size Size of the data payload in bytes.
+        /// \param process   Callback invoked on response: returns the resulting datagram state.
+        /// \param error     Callback invoked when the datagram was lost or its processing failed.
+        virtual void addDatagram(enum Command command, uint32_t address, void const* data, uint16_t data_size,
+                                 std::function<DatagramState(DatagramHeader const*, uint8_t const* data, uint16_t wkc)> const& process,
+                                 std::function<void(DatagramState const& state)> const& error) = 0;
+
+        /// \brief Convenience overload: queue a datagram with a typed payload.
+        template<typename T>
+        void addDatagram(enum Command command, uint32_t address, T const& data,
+                         std::function<DatagramState(DatagramHeader const*, uint8_t const* data, uint16_t wkc)> const& process,
+                         std::function<void(DatagramState const& state)> const& error)
+        {
+            addDatagram(command, address, &data, sizeof(data), process, error);
+        }
+
+        /// \brief Flush any remaining queued datagrams onto the wire.
+        virtual void finalizeDatagrams() = 0;
+
+        /// \brief Send all queued datagrams, read responses, and invoke callbacks.
+        virtual void processDatagrams() = 0;
+
+        /// \brief Set the read timeout for the underlying sockets.
+        virtual void setTimeout(nanoseconds const& timeout) = 0;
+
+        /// \brief Check if redundancy is needed to handle the current bus. If yes it means to system is already degraded.
+        virtual void checkRedundancyNeeded() = 0;
+
+        /// \brief Register a callback to be called when an EtherCAT IRQ event is detected.
+        virtual void attachEcatEventCallback(enum EcatEvent event, std::function<void()> callback) = 0;
+    };
+}
+
+#endif
--- a/lib/master/include/kickcat/Link.h
+++ b/lib/master/include/kickcat/Link.h
@@ -4,13 +4,13 @@
 #include <array>
 #include <functional>
 
-#include "kickcat/Frame.h"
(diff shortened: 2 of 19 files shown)


lib/master/include/kickcat/AbstractLink.h
```diff
@@ -0,0 +1,63 @@
...
+        /// \param data_size Size of the data payload in bytes.
+        /// \param process   Callback invoked on response: returns the resulting datagram state.
+        /// \param error     Callback invoked when the datagram was lost or its processing failed.
+        virtual void addDatagram(enum Command command, uint32_t address, void const* data, uint16_t data_size,
+                                 std::function<DatagramState(DatagramHeader const*, uint8_t const* data, uint16_t wkc)> const& process,
+                                 std::function<void(DatagramState const& state)> const& error) = 0;
+
+        /// \brief Convenience overload: queue a datagram with a typed payload.
+        template<typename T>
+        void addDatagram(enum Command command, uint32_t address, T const& data,
```
trns1997, line 39: I do not really understand why this method is required

lib/master/include/kickcat/Link.h
```diff
@@ -23,33 +23,23 @@ namespace kickcat
...
+        ~Link() override = default;
 
-        /// \brief   Helper for trivial access (i.e. most of the init bus frames)
-        ///
-        /// \details Since this method is only used for non real time operation, the redundancy mechanism used is slower
-        ///          but guaranty slave order access, ie for setAddresses().
-        void writeThenRead(Frame& frame) ;
+        void writeThenRead(Frame& frame) override;
 
+        using AbstractLink::addDatagram;
```
trns1997, line 30: why is this forward declaration required? The template `addDatagram` calls the override method no?

tools/eeprom.cc
```diff
@@ -2,6 +2,7 @@
 // Current state: POC, can only dump eeprom file.
 
 #include "kickcat/Bus.h"
+#include "kickcat/Link.h"
```
trns1997, line 5: this include seems useless to me

tools/scanTopology.cc
```diff
@@ -1,5 +1,6 @@
 #include "kickcat/Bus.h"
 #include "kickcat/Diagnostics.h"
+#include "kickcat/Link.h"
```
trns1997, line 3: is this include necessary?

unit/src/bus-t.cc
```diff
@@ -667,19 +572,10 @@ TEST_F(BusTest, send_get_DL_status)
...
-    checkSendFrame(Command::BWR, uint16_t(EcatEvent::DL_STATUS | EcatEvent::AL_STATUS));
-    handleReplyWriteThenRead();
-
-    checkSendFrame(Command::BWR, uint16_t(EcatEvent::AL_STATUS));
-    handleReplyWriteThenRead();
-
-    checkSendFrame(Command::BWR, uint16_t(0));
-    handleReplyWriteThenRead();
+    mock_link->handleWriteThenRead(1);
+    mock_link->handleWriteThenRead(1);
```
trns1997: is this test equivalent, now we check only 4 write then read are done there is not validity of the masks like before.

unit/src/bus-t.cc
```diff
@@ -23,129 +16,69 @@ struct SDOAnswer
...
     void initBus(milliseconds watchdog = 100ms)
     {
-        InSequence s;
-
         detectAndReset();
 
-        // PDIO Watchdog
-        uint16_t watchdogTimeCheck = static_cast<uint16_t>(watchdog / 100us);
-
-        std::vector<DatagramCheck<uint16_t>> expecteds_1(1, {Command::BWR, uint16_t(0x09C2)});
```
trns1997, line 118: we no longer do this?

unit/src/bus-t.cc
```diff
@@ -23,129 +16,69 @@ struct SDOAnswer
...
-        io_nominal->checkSendFrame(expecteds_1);
-        handleReplyWriteThenRead();
-
-        std::vector<DatagramCheck<uint16_t>> expecteds_2(1, {Command::BWR, watchdogTimeCheck});
```
trns1997, line 122: same here we no longer check the contents?

unit/src/bus-t.cc
```diff
@@ -314,54 +239,49 @@ TEST_F(BusTest, error_counters)
...
     }
 
+    // test logical write
     int64_t logical_write = 0x0706050403020100;
     std::memcpy(slave.output.data, &logical_write, sizeof(int64_t));
-    std::vector<DatagramCheck<int64_t>> expecteds(1, {Command::LWR, logical_write});
-    io_nominal->checkSendFrame(expecteds);
-
-    io_nominal->handleReply<int64_t>({logical_write});
+    mock_link->handleProcess<int64_t>(logical_write, 1);
```
trns1997: we no longer verify the written payloads?

unit/mocks/Link.h
```diff
@@ -0,0 +1,146 @@
...
+            if (data != nullptr && data_size > 0)
+            {
+                dg.data.resize(data_size);
+                std::memcpy(dg.data.data(), data, data_size);
+            }
+            pending_datagrams_.push_back(std::move(dg));
+        }
+
+        template<typename T>
+        void handleProcess(T const& reply_data, uint16_t wkc = 1)
```
trns1997: i think we should add an optional argument to store `expected_cmd` which can then be asserted equal in `processDatagrams` to have the same effect as `io_nominal->checkSendFrame` right now we just process blindly (which can be ok if we say it is ok). Here is what the mock socket does: https://github.com/leducp/KickCAT/blob/2254a1adae432013272522e48a8cc5187845ecf6/unit/mocks/Sockets.h#L45

lib/master/include/kickcat/AbstractLink.h
leducp: It is just a helper to avoid manipulating pointers

lib/master/include/kickcat/Link.h
leducp: But not the helper :)

tools/eeprom.cc
leducp: It is needed to create the concrete Link (before it was provided by Bus.h but now it is only the abstract one)

tools/scanTopology.cc
leducp: same:
It is needed to create the concrete Link (before it was provided by Bus.h but now it is only the abstract one)

unit/src/bus-t.cc
leducp: On it

lib/master/include/kickcat/Link.h
trns1997: Ah right of course, gotta expose it :)

unit/src/bus-t.cc
leducp: not good no: on it

unit/src/bus-t.cc
leducp: we do need it, especially with the link as mock it should be easier to test

unit/mocks/Link.h
leducp: On it: we do not want to degrade the checking behavior

tools/eeprom.cc
trns1997: ah right sorry on github it didnt open all the content did not see link in the code

tools/scanTopology.cc
trns1997: idem

PR #304, opened 2026-03-02
refactor map pdo method and move it to bus

### LINKED ISSUE #302: Expose a global Map PDO method

trns1997:
Closes #302

- [x] Tested on Freedom K64F

### DIFF
--- a/examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
+++ b/examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
@@ -125,10 +125,10 @@ int main(int argc, char* argv[])
         printf("\n=== Configuring PDO Mappings (Accelerometer Only) ===\n");
 
         // Map TxPDO (slave -> master): Accelerometer data
-        bus.mapPDO(bus.slaves().at(0), 0x1A00, pdo::tx_mapping, pdo::tx_mapping_count, 0x1C13);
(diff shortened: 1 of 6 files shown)


examples/master/freedom-k64f/freedom_k64f_accel_map_example.cc
```diff
@@ -150,10 +125,10 @@ int main(int argc, char* argv[])
         printf("\n=== Configuring PDO Mappings (Accelerometer Only) ===\n");
 
         // Map TxPDO (slave -> master): Accelerometer data
-        mapPDO(0, 0x1A00, pdo::tx_mapping, pdo::tx_mapping_count, 0x1C13);
+        bus.mapPDO(bus.slaves().at(0), 0x1A00, pdo::tx_mapping, pdo::tx_mapping_count, 0x1C13);
```
leducp: I would prefer either a free function or a function associated to the slave: bus class is too big already and it is only needed here to do the SDO (which can be done through the slave mailbox directly, then run through a few bus call)

trns1997: free function in the bus hows that? Our you want in in a utils fiile?

trns1997: or even better in the CoE file?

PR #312, opened 2026-03-04
fix binding build and add checks in CI

### LINKED ISSUE #313: Fix failing build python binding
- Fix build issue
- Add stage to CI to prevent future regression

trns1997:
Closes #313

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: PR build
+name: CI pipeline
 
 on:
   push:
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -4,7 +4,7 @@ name: Quality
 # access to secrets
 on:
   workflow_run:
-    workflows: ["PR build"]
+    workflows: ["CI pipeline"]
     types:
       - completed
 


.github/workflows/build.yml
```diff
@@ -58,6 +58,14 @@ jobs:
       - name: Build
         run: cmake --build build --config ${{env.BUILD_TYPE}}
 
+      - name: Build Python Bindings
```
leducp: This should be unnecessary as there is another job building the wheels: we should probably do a better workflow:
step 1: 
- build wheel
- build
step 2:
- unit tests
- sim tests (with python scripts and bin examples)
step 3: 
- release (on tag)

trns1997: ok i can rework this, these stages seem logical and that way dont have redundant lib build, reusable approach

trns1997: @leducp checkout out the pipeline now based on your recommendation: https://github.com/leducp/KickCAT/actions/runs/22672154802?pr=312

If you think it looks good then i will rename build.yml to ci.yml as it does a lot more than build now. I needed to move the python wheel workflow into this file otherwise impossible to create a dependency easily with github ci :(

leducp: Seems good. Yes Github CI is a pain in the ass...

PR #324, opened 2026-03-08
Add support for AF_XDP socket (expect lower latency and slighty lower CPU load)

leducp:


lib/master/src/helpers.cc
```diff
@@ -74,47 +77,40 @@ namespace kickcat
...
+            if (ifname.compare(0, XDP_PREFIX_LEN, XDP_PREFIX) == 0)
+            {
+                ifname = ifname.substr(XDP_PREFIX_LEN);
+                return std::make_shared<XdpSocket>();
+            }
+#endif
+            return std::make_shared<Socket>();
+        };
+
+        nominal = createSocket(nominal_if, "tap_nominal");
```
trns1997, line 109: i find this a bit confusing to read. Because the `tap_name` is  only ever used if a tap interface is used. Just a remark not super important tbf

leducp: I understand but this seems more clean that creating a constant somewhere no?

lib/src/OS/Linux/XdpSocket.cc
```diff
@@ -0,0 +1,330 @@
...
+        {
+            munmap(umem_area_, umem_size);
+            umem_area_ = nullptr;
+            THROW_SYSTEM_ERROR("xsk_umem__create()");
+        }
+
+        struct xsk_socket_config xsk_cfg = {};
+        xsk_cfg.rx_size      = RX_RING_SIZE;
+        xsk_cfg.tx_size      = TX_RING_SIZE;
+        xsk_cfg.bind_flags   = XDP_USE_NEED_WAKEUP;
```
trns1997, line 147: you set this flag so, we are responsible poll and manually do the syscall see: https://docs.ebpf.io/linux/concepts/af_xdp/#options-variations-and-exceptions but i do not see the `recvfrom` syscall anywhere is this embedded somewhere. I am guessing this should be done in `refillFillRing` but i am not entirely sure what it should look like. I just want to make sure we take this into account that the process is responsible and should explicitly check.

lib/src/OS/Linux/XdpSocket.cc
```diff
@@ -0,0 +1,330 @@
...
+        uint32_t idx = 0;
+        uint32_t completed = xsk_ring_cons__peek(&comp_ring_, COMP_RING_SIZE, &idx);
+        if (completed == 0)
+        {
+            return;
+        }
+
+        for (uint32_t i = 0; i < completed; ++i)
+        {
+            freeFrame(*xsk_ring_cons__comp_addr(&comp_ring_, idx + i));
```
trns1997, line 216: over here https://github.com/leducp/KickCAT/pull/324/changes#diff-f833bdc5c5a5a35133303625c6ef9c40ee89f5c906657e6f4971ed0fdb6db9a2R277 after free frame you call `refillFillRing` shouldn't we do the same here?

lib/src/OS/Linux/XdpSocket.cc
leducp: and this is a bug... fixed

trns1997: Don't need to do the same with `sendto`?

leducp: There is sendto already in write

lib/src/OS/Linux/XdpSocket.cc
leducp: it is done at the end to work in batch

PR #351, opened 2026-03-24
mailbox check through PDOs

### LINKED ISSUE #56: Handle mailbox checking through FMMU
Allow the client to enable mailbox checking through FMMU.

leducp:
Closes #56 

Add a config that enable the usage of remaining FMMUs to fetch the mailbox status directly at the end of the PDOs/Process Image provided that the slave has enough FMMUs available (i.e. lan9252 have only 3 FMMUs, so all checks cannot be enable).

Note that for now the FMMus selection is static:
FMMU0/1 -> PDOs
FMMU2/3 -> mbx checks

### DIFF
--- a/lib/master/include/kickcat/Bus.h
+++ b/lib/master/include/kickcat/Bus.h
@@ -10,6 +10,21 @@
 
 namespace kickcat
 {
+    enum MailboxStatusFMMU : uint8_t
+    {
+        NONE        = 0,
+        READ_CHECK  = (1 << 0),
+        WRITE_CHECK = (1 << 1),
+    };
+    constexpr MailboxStatusFMMU operator|(MailboxStatusFMMU a, MailboxStatusFMMU b)
+    {
+        return static_cast<MailboxStatusFMMU>(static_cast<uint8_t>(a) | static_cast<uint8_t>(b));
+    }
+    constexpr bool operator&(MailboxStatusFMMU a, MailboxStatusFMMU b)
+    {
+        return (static_cast<uint8_t>(a) & static_cast<uint8_t>(b)) != 0;
+    }
+
     class Bus
     {
     public:
@@ -53,6 +68,17 @@ namespace kickcat
         // wait for all slaves to reached a state
         // background_task may be used to keep updated PDO while waiting for a particular state.
         void waitForState(State request, nanoseconds timeout, std::function<void()> background_task = [](){});
+
+        /// \brief Select and validate the preferred mailbox status check mode.
+        /// \details Must be called during PRE_OP, before createMapping(). Validates that all
+        ///          mailbox-capable slaves have enough FMMUs for the requested mode. Throws on error.
+        ///          When set, createMapping() transparently integrates bit-wise FMMU mapping so that
+        ///          mailbox status flags (can_read / can_write) are updated in the LRD/LRW callbacks.
+        /// \param mode  READ_CHECK, WRITE_CHECK, or both OR'd together
+        void configureMailboxStatusCheck(MailboxStatusFMMU mode);
+
+        /// \return the currently configured mailbox status FMMU mode
+        MailboxStatusFMMU mailboxStatusFMMUMode() const { return mailbox_status_fmmu_; }
 
         // create the mapping between slaves PI and client buffer
         // if OK, set the bus to SAFE_OP state
@@ -163,6 +189,7 @@ namespace kickcat
         void detectMapping();
         void readMappedPDO(Slave& slave, uint16_t index);
         void configureFMMUs();
+        void configureMailboxFMMUs();
 
         // DC helpers
         void fetchReceivedTimes();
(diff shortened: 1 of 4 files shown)


lib/master/include/kickcat/Bus.h
```diff
@@ -183,12 +210,23 @@ namespace kickcat
...
+
         struct PIFrame
         {
             uint32_t address;               // logical address
-            int32_t size;                   // frame size
+            int32_t  logical_size;          // full logical address range (PDO + mailbox status bits)
+            int32_t  pdo_size;             // process data only (used by LWR)
             std::vector<blockIO> inputs;    // slave to master
             std::vector<blockIO> outputs;
+            std::vector<MailboxStatusEntry> mailbox_read_status;
```
trns1997, line 227: complete the comments helpful

lib/master/src/Bus.cc
```diff
@@ -507,10 +542,84 @@ namespace kickcat
...
+            {
+                auto& frame = pi_frames_[fi];
+                auto const& mbx_slaves = frame_mbx_slaves[fi];
+                if (mbx_slaves.empty())
+                {
+                    continue;
+                }
+
+                uint32_t status_offset = static_cast<uint32_t>(frame.pdo_size);
+                constexpr uint32_t FMMU_BIT_SEPARATION = 3;
```
trns1997:
```suggestion
                constexpr uint32_t FMMU_BYTE_SEPARATION = 3;
```

Shouldn't this be `byte` and not `bit`?

lib/master/src/Bus.cc
```diff
@@ -507,10 +542,84 @@ namespace kickcat
...
+                status_offset += FMMU_BIT_SEPARATION;
```
trns1997: A small comment why we always add 3 bytes even if read and write checks are not active. Padding?

lib/master/src/Bus.cc
```diff
@@ -507,10 +542,84 @@ namespace kickcat
...
+                    {
+                        frame.mailbox_read_status.push_back({
+                            status_offset + static_cast<uint32_t>(i / 8),
+                            static_cast<uint8_t>(i % 8),
+                            mbx_slaves[i]
+                        });
+                    }
+                    uint32_t read_bytes = (static_cast<uint32_t>(mbx_slaves.size()) + 7) / 8;
+
+                    if (mailbox_status_fmmu_ & MailboxStatusFMMU::WRITE_CHECK)
```
trns1997: Not a fan of having this silent dependency. It might be clearer maybe to compute frame layout first and then write the frame? Not sure what you think?

lib/master/include/kickcat/Bus.h
```diff
@@ -183,12 +210,23 @@ namespace kickcat
...
+            std::vector<MailboxStatusEntry> mailbox_write_status;
+            uint16_t mailbox_status_wkc_read_adjust{0};
```
trns1997, line 231: small comment i am not sure i understand this wkc

lib/master/src/Bus.cc
```diff
@@ -725,6 +860,74 @@ namespace kickcat
...
+                    bus_info("Mailbox FMMU slave %04x - FMMU%d - logical 0x%08x bit %d - physical 0x%04x bit 3\n",
+                        entry.slave->address, fmmu_reg_offset / 0x10,
+                        fmmu.logical_address, entry.bit_position, physical_address);
+                }
+            };
+
+            // FMMU0 and FMMU1 are reserved for PDO (output and input).
+            // FMMU2 is used by the first enabled mailbox check mode.
+            // FMMU3 is used by the second one, only when both modes are active.
+            if (mailbox_status_fmmu_ & MailboxStatusFMMU::READ_CHECK)
```
trns1997, line 919:
```
constexpr uint16_t FMMU2_OFFSET  = 0x20;
constexpr uint16_t FMMU_REG_SIZE = 0x10;

uint16_t fmmu_offset = FMMU2_OFFSET;

if (mailbox_status_fmmu_ & MailboxStatusFMMU::READ_CHECK)
{
    configureBitFMMU(frame.mailbox_read_status,
        static_cast<uint16_t>(reg::SYNC_MANAGER_1 + reg::SM_STATS),
        fmmu_offset);
    fmmu_offset += FMMU_REG_SIZE;
}

if (mailbox_status_fmmu_ & MailboxStatusFMMU::WRITE_CHECK)
{
    configureBitFMMU(frame.mailbox_write_status,
        static_cast<uint16_t>(reg::SYNC_MANAGER_0 + reg::SM_STATS),
        fmmu_offset);
    fmmu_offset += FMMU_REG_SIZE;
}
```

Slight rework what do you think?

lib/master/src/Bus.cc
leducp: From the doc (section 1, chapter 7):

The logical address ranges of two FMMUs of the same direction (read or write) in one ESC must be
separated by at least 3 logical bytes not configured by any FMMU of the same type, if one of the
FMMUs or both use bit-wise mapping (logical start bit ≠ 0, logical stop bit ≠ 7, or physical start bit ≠ 0).

leducp: So yes this is a mistake.

lib/master/src/Bus.cc
leducp: Here there is always at least one check enable because we are in the if() condition that specifically check this?
 BTW I update the code because I misread the doc and we also need a systematic 3 bytes gap between read check and write check

lib/master/src/Bus.cc
leducp: which dependency are you talking about?

lib/master/src/Bus.cc
leducp: Done

lib/master/src/Bus.cc
leducp: Comment added

lib/master/src/Bus.cc
trns1997: the `WRITE_CHECK` nested in the `READ_CHECK`, i get that they are intimately linked, but in the case we add another flag between read and write (not likely) this coupling would break. But again not sure we will have other flags in the near future

trns1997 (approved):

trns1997 (approved):

PR #366, opened 2026-04-01
Eeprom editor

### LINKED ISSUE #277: EEPROM/ESI editor
- GUI app to modify or generate ESI XML file
- Tool script `esi2eeprom` or maybe directly in the GUI app (might be better)

leducp:
Closes #277

### DIFF
--- a/tools/eeprom_editor/App.cc
+++ b/tools/eeprom_editor/App.cc
@@ -1,37 +1,35 @@
-#include "App.h"
-
 #include <cstdio>
 #include <fstream>
 #include <iterator>
 
-#include "imgui.h"
-#include "portable-file-dialogs.h"
+#include <imgui.h>
+#include <portable-file-dialogs.h>
 
 #include "kickcat/EEPROM/EEPROM_factory.h"
 
+#include "App.h"
+#include "Editors.h"
+
 namespace kickcat::eeprom_editor
 {
-    namespace
+    constexpr float SIDEBAR_WIDTH = 220.0f;
+
+    struct CategoryInfo
     {
-        constexpr float SIDEBAR_WIDTH = 220.0f;
+        Category id;
+        char const* label;
+    };
 
-        struct CategoryInfo
-        {
-            Category id;
-            char const* label;
-        };
-
-        constexpr CategoryInfo CATEGORIES[] =
-        {
-            { Category::Info,         "Info (Header)"  },
-            { Category::Strings,      "Strings"        },
-            { Category::General,      "General"        },
-            { Category::SyncManagers, "Sync Managers"  },
-            { Category::FMMU,         "FMMU"           },
-            { Category::TxPDO,        "TxPDO"          },
-            { Category::RxPDO,        "RxPDO"          },
-        };
-    }
+    constexpr CategoryInfo CATEGORIES[] =
+    {
+        { Category::Info,         "Info (Header)"  },
+        { Category::Strings,      "Strings"        },
+        { Category::General,      "General"        },
+        { Category::SyncManagers, "Sync Managers"  },
+        { Category::FMMU,         "FMMU"           },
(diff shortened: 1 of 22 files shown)


trns1997 (dismissed):

tools/eeprom_editor/App.cc
```diff
@@ -0,0 +1,979 @@
...
+        });
+    }
+
+    // ── Dialog rendering ───────────────────────────────────────────────
+
+    void App::renderConnectDialog()
+    {
+        if (show_connect_dialog_)
+        {
+            ImGui::OpenPopup("Connect to EtherCAT Network");
```
trns1997: https://github.com/ocornut/imgui/discussions/3773#discussioncomment-327082

Here they suggest that calling `ImGui::OpenPopup()` on every frame is not good. We should consider have maybe a `popup_opened` flag to avoid calling it again and again

tools/eeprom_editor/App.cc
```diff
@@ -0,0 +1,979 @@
...
+        {
+            ImGui::TextColored(COLOR_RED,
+                               "CRC 0x%02X != 0x%02X", actual_crc, expected_crc);
+        }
+
+        statusSeparator();
+
+        if (isBusy())
+        {
+            std::lock_guard lock(worker_status_mutex_);
```
trns1997: shouldn't this lock be outside the `if` as the `else if` accesses a `worker_error_` which everywhere is else if access with the mutex?

tools/eeprom_editor/App.cc
```diff
@@ -0,0 +1,979 @@
...
+                    saveFile();
+                }
+                if (ImGui::MenuItem("Save As...", "Ctrl+Shift+S"))
+                {
+                    saveFileAs();
+                }
+                ImGui::Separator();
+                if (ImGui::MenuItem("Quit", "Ctrl+Q"))
+                {
+                    std::exit(0);
```
trns1997, line 240: https://stackoverflow.com/questions/7054685/are-destructors-run-when-calling-exit

doesn'nt imgui have its own close method thats potentially safer?

tools/eeprom_editor/App.cc
```diff
@@ -0,0 +1,979 @@
...
+    void App::loadFromSlave(int slave_index)
+    {
+        try
+        {
+            auto& slave = bus_->slaves().at(slave_index);
+            auto data = slave.sii.serialize();
+
+            sii_ = eeprom::SII{};
+            sii_.parse(data);
+            serialized_ = sii_.serialize();
```
trns1997, line 587: umm isn't this redundant ? `sii_ = slave.sii` and `serialized = data`? why is reparsing necessary here?

tools/eeprom_editor/App.cc
```diff
@@ -0,0 +1,979 @@
...
+
+                ImGui::TableNextColumn();
+                ImGui::Text("%s", resolveString(slave.sii.strings, slave.sii.general.device_name_id));
+            }
+            ImGui::EndTable();
+        }
+        return double_clicked;
+    }
+
+    void App::renderSlaveDialog()
```
trns1997: I think the Dialog methods like `renderConnectDialog` and `renderSlaveDialog` share a lot of similarities, i think the dialog logic can be abstracted to create a generic helper. Could be useful for future dialogs as well.

tools/eeprom_editor/App.cc
leducp: good catch: too many refactors ^^'

tools/eeprom_editor/App.cc
leducp: I don't think so? In this case I think it is safe: we are closing the app so any unsaved work will be lost as expected.

trns1997 (approved):

PR #376, opened 2026-04-15
Continue ETG.1510 work

leducp:


lib/src/Mailbox.cc
```diff
@@ -434,6 +434,40 @@ namespace kickcat::mailbox::response
...
+        {
+            // Empty or malformed reply (the dictionary or a factory produced nothing routable).
+            return nullptr;
+        }
+
+        // Wrap the reply and drive it to SUCCESS by tagging its header with the gateway-masked
+        // address and feeding it back through GatewayMessage::process(). process() then restores
+        // the original (master) address on the delivered frame and flips status to SUCCESS, which
+        // is what Gateway::processPendingRequests() looks for.
+        uint16_t mailbox_size = static_cast<uint16_t>(
```
trns1997, line 459: I do not get why you need `max()` because `process()` calls `data_.resize(size)` using the reply's own `header->len`. Can this simply not just be `uint16_t mailbox_size = static_cast<uint16_t>(reply.size());`?

lib/master/src/MasterOD.cc
```diff
@@ -72,27 +72,57 @@ namespace kickcat
...
+            bit_offset += 32;
+            CoE::addEntry<uint32_t>(obj, 6,  32,          bit_offset, CoE::Access::READ, CoE::DataType::UNSIGNED32,     "Product Code",                sii.info.product_code);
+            bit_offset += 32;
+            CoE::addEntry<uint32_t>(obj, 7,  32,          bit_offset, CoE::Access::READ, CoE::DataType::UNSIGNED32,     "Revision Number",             sii.info.revision_number);
+            bit_offset += 32;
+            CoE::addEntry<uint32_t>(obj, 8,  32,          bit_offset, CoE::Access::READ, CoE::DataType::UNSIGNED32,     "Serial Number",               sii.info.serial_number);
+            bit_offset += 32;
+            CoE::addEntry<uint16_t>(obj, 33, 16,          bit_offset, CoE::Access::READ, CoE::DataType::UNSIGNED16,     "Mailbox Out Size",            sii.info.standard_recv_mbx_size);
+            bit_offset += 16;
+            CoE::addEntry<uint16_t>(obj, 34, 16,          bit_offset, CoE::Access::READ, CoE::DataType::UNSIGNED16,     "Mailbox In Size",             sii.info.standard_send_mbx_size);
```
trns1997, line 107: sub index `35` missing same comment as subindex `02`?

lib/src/Mailbox.cc
leducp: This logic was inherited from  the slave path, but it is wasteful: fixed by adding a new ctor to wrap reply ready to be shipped

lib/master/src/MasterOD.cc
leducp: This was an error: fixed (and same for 2 BTW, it was in the table ^^')

trns1997 (approved):

PR #381, opened 2026-04-23
Add C++ automatic formatting

windelbouwman:
Initial version of clang format file.


trns1997:
Can we run this on a particular folder (ideally isolate a section of the code) to get an idea on the impact on the current code base?

windelbouwman:
> Can we run this on a particular folder (ideally isolate a section of the code) to get an idea on the impact on the current code base?

That's a great idea. I formatted the slave examples, and added this as a CI check step.

examples/slave/nuttx/lan9252/arduino-due/main.cc
```diff
@@ -13,10 +13,10 @@
 using namespace kickcat;
 
 
-int main(int argc, char *argv[])
+int main(int argc, char* argv[])
```
trns1997, line 16: I think we want it right aligned, but I forgot what @leducp preference is :/

trns1997:
> > Can we run this on a particular folder (ideally isolate a section of the code) to get an idea on the impact on the current code base?
> 
> That's a great idea. I formatted the slave examples, and added this as a CI check step.

Honestly, I think it looks good for the files chosen so far. I think worth to look into couple files in the lib, as we have switch cases that are formatted in a way we deem human readble, I am curious to see what uncrustify does there.

PR #382, opened 2026-04-23
Refactor timer

leducp:
* remove useless sync with mutex+cond_var, 
* start is now a wrapper on init for retrocompatibility (there is no more notion of 'stop' so 'start' is strange


lib/include/kickcat/OS/Timer.h
```diff
@@ -17,19 +14,17 @@ namespace kickcat
...
+        /// \details Aligns the next deadline to sync_point + N*period so that it lies in the future.
+        ///          Safe to call multiple times.
+        void init(nanoseconds sync_point = since_epoch());
 
-        ///  Stop the timer.
-        void stop();
-
-        ///  Get timer status
-        bool is_stopped() const;
+        /// \deprecated Use init() instead.
```
trns1997: Do you want to try use this https://en.cppreference.com/cpp/language/attributes/deprecated
?

trns1997 (approved):

PR #384, opened 2026-04-23
Modify the network simulator. Users have 2 ways to input the desired number of EtherCAT slaves on the network.

phamnhatha090805:
I did some changes in the network_simulator.cc so now users have 2 ways to give the amount of EtherCAT slaves that they want on the network. Users can say `-i <network_interface> -n <number_of_slave> -s <config_file.json>` or user can also do the old way `-i <network_interface> -s <config_file_1.json config_file_2.json ...>`. I don't know if this is something you guys want or not so you can review this little changes. Personally, I found this is more convenient if I want to have 100 virtual EtherCAT devices on the simulated network.

### DIFF
--- a/simulation/network_simulator.cc
+++ b/simulation/network_simulator.cc
@@ -82,7 +82,7 @@ int main(int argc, char* argv[])
         expanded_slave_configs = slave_configs;
     }
 
-        if (slave_configs.empty())
+	if (slave_configs.empty())
     {
         std::cerr << "No slave configuration files provided" << std::endl;
         std::cerr << program;
--- a/simulation/network_simulator.cc
+++ b/simulation/network_simulator.cc
@@ -30,6 +30,13 @@ int main(int argc, char* argv[])
         .required()
         .store_into(interface);
 
+    int slave_number = 0;
+    program.add_argument("-n", "--count")
+        .help("Number of slaves to simulate")
+        .default_value(0)
+        .scan<'i', int>()
+        .store_into(slave_number);
+
     std::vector<std::string> slave_configs;
     program.add_argument("-s", "--slaves")
         .help("JSON configuration files for slaves")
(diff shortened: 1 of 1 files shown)


simulation/network_simulator.cc
```diff
@@ -54,7 +61,35 @@ int main(int argc, char* argv[])
...
+        {
+            expanded_slave_configs.push_back(slave_configs[0]);
+        }
+    }
+    else
+    {
+        expanded_slave_configs = slave_configs;
+    }
+
+        if (slave_configs.empty())
```
leducp: alignment

leducp:
That's a good idea :)

trns1997:
I think the idea makes sense. Being able to quickly spin up something like 100 virtual EtherCAT slaves is genuinely useful for testing and scaling scenarios.

That said, I’d push for a single, consistent way of defining the network rather than supporting multiple input modes. Right now we have:

- "-n <number_of_slaves>" (implicit generation)
- multiple JSON configs (explicit definition)

Maintaining both introduces ambiguity and increases the surface for bugs and inconsistencies over time.

A cleaner approach would be to move everything into the JSON config and make it the single source of truth. For example, the slave count (or generation rules) could live inside the JSON itself. That way:

- the configuration is self-contained
- the interface stays simple (one input method)

Also, I think we’re currently mixing concepts a bit between ESI and ENI:

- ESI (EtherCAT SubDevice Information): describes what a slave can do (capabilities, PDOs, objects, etc.), typically provided by the vendor.
- ENI (EtherCAT Network Information): describes how slaves are actually instantiated and connected in a specific network (topology, mappings, runtime configuration).

Right now, the simulator interface seems to blur those two responsibilities. Ideally:

- the JSON should act more like an ENI (network definition)
- reusable slave descriptions should stay closer to ESI-like definitions

Not blocking for this PR, but I’d see your change as a good stepping stone toward a more standard, EtherCAT-aligned configuration model.

What do you guys thing?

leducp:
Let's start simple: I vote for merging this PR first to support a real need. But you are right: we need to add a ENI support to do exactly that: create a whole network.
We probably need ot support the offical ENI format, a tool to edit it easily and maybe a custom format that is easier to tweak manually.

leducp:
In // I'm starting to create a proper integration test that leverage the emulator/tapsocket in the CI to catch regression.

phamnhatha090805:
> Let's start simple: I vote for merging this PR first to support a real need. But you are right: we need to add a ENI support to do exactly that: create a whole network. We probably need ot support the offical ENI format, a tool to edit it easily and maybe a custom format that is easier to tweak manually.

About the ENI and official Beckhoff ESI support. I have made some changes so user can simulate network with official Beckhoff EtherCAT modules. I want to run some test and clean the commits a bit before make a pull request.

trns1997 (approved):

PR #386, opened 2026-04-24
Improve the network_simulator and EsiParser

phamnhatha090805:
This modification is making the KickCAT more compliance with the official Beckhoff EtherCAT module. So the XML file from Beckhoff can contain several EtherCAT devices (e.g. EL1xxx.xml) but the EEPROM BIN only contains 1 device. With these changes, the program will get the vendor_id, product_code, and revision_number from the EEPROM BIN. From that, the program go look into the xml file until it find the matching vendor_id, product_code, and revision_number instead of finding the first elements. 
Also I agree with the ENI vs ESI seperation because for example the EL1xxx modules need a EtherCAT bus coupler infront of it. It's more convenient to have a .json file that contains whole network instead of `-s <EtherCAT coupler (EK1xxx)>.json <EL1xxx>.json`

### DIFF
--- a/examples/master/load_esi/load_esi.cc
+++ b/examples/master/load_esi/load_esi.cc
@@ -28,7 +28,7 @@ int main(int argc, char const* argv[])
     CoE::EsiParser parser;
 
     nanoseconds t1 = since_epoch();
-    CoE::Dictionary coe_dict = parser.loadFile(esi_file);
+    CoE::Dictionary coe_dict = parser.loadFirstDictionaryFromFile(esi_file);
     nanoseconds t2 = since_epoch();
 
     // dangerous lack of error checking.
--- a/lib/include/kickcat/CoE/EsiParser.h
+++ b/lib/include/kickcat/CoE/EsiParser.h
@@ -14,17 +14,18 @@ namespace kickcat::CoE
         EsiParser() = default;
         ~EsiParser() = default;
 
-        CoE::Dictionary loadFile  (std::string const& file);
-        CoE::Dictionary loadString(std::string const& xml);
+        Dictionary loadFirstDictionaryFromFile  (std::string const& file);
+        std::vector<Device> loadDevicesFromFile  (std::string const& file);
+        std::vector<Device> loadString(std::string const& xml);
 
         char const* vendor() const  { return vendor_->FirstChildElement("Name")->GetText();       }
         char const* profile() const { return profile_->FirstChildElement("ProfileNo")->GetText(); }
 
     private:
+    
         template<typename T>
-        T toNumber(tinyxml2::XMLElement* node)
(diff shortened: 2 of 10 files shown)


leducp:
I though ESI should only handle one slave but ENI multiple as it describes a network? If yes we need a deidcated parser (ENIParser)



Answering to myself: yes an ESI can have multiple slaves

lib/slave/src/ESC/EmulatedESC.cc
```diff
@@ -69,10 +69,25 @@ namespace kickcat
...
+        // 2. Station Alias (Word 4)
+        // Note: If you use Word 4 for Alias, ensure your BIN file actually has 
+        // an alias there, otherwise it may interfere with identity reading logic.
+        memory_.station_alias = eeprom_[4];
+
+        // 3. MANDATORY IDENTITY OFFSETS (Words 8 through 13)
+        // In many Beckhoff BIN files, the Identity section starts at Word 8 (Byte 16)
+        
+        // Vendor ID (Word 8 & 9)
+        vendor_id_ = (static_cast<uint32_t>(eeprom_[9]) << 16) | eeprom_[8];
```
leducp, line 84: I wonder if it is better to use the SII Parser to extract the data needed here (it can also verify the eeprom coherency).

examples/master/load_esi/load_esi.cc
```diff
@@ -28,7 +28,7 @@ int main(int argc, char const* argv[])
     CoE::EsiParser parser;
 
     nanoseconds t1 = since_epoch();
-    CoE::Dictionary coe_dict = parser.loadFile(esi_file);
+    CoE::Dictionary coe_dict = parser.loadFirstDictionaryFromFile(esi_file);
```
leducp, line 31: Is it to handle legacy call?  If yes I think it is better to just have one API that retrieve the vector of dic and let each example get themself the first one (they are examples after all :) )

tools/od_generator.cc
```diff
@@ -147,7 +147,7 @@ namespace kickcat
     CoE::Dictionary loadOD(std::string esiFileName)
```
leducp, line 147: Unrelated to you PR but could you convert this call to loadOD(std::string const& esiFileName)` ?

lib/include/kickcat/CoE/EsiParser.h
```diff
@@ -70,4 +80,4 @@ namespace kickcat::CoE
     };
 }
 
-#endif
+#endif
```
trns1997, line 83:
```suggestion
#endif

```

tools/od_generator.cc
phamnhatha090805: Sure!

PR #387, opened 2026-05-07
ESI parser compliance - step 1

leducp:
This is the first PR of a long run work to reach 100% compatibility with ETG standard

* Move ESI Parser in a dedicated folder and namespace
* Handle multiple devices in the same file


lib/src/ESI/Parser.cc
```diff
@@ -0,0 +1,667 @@
...
+        bitoff = toNumber<uint16_t>(node_bitoff);
+    }
+
+    return {type, bitlen, bitoff};
+}
+
+
+XMLNode* Parser::findNodeType(XMLNode* node)
+{
+    std::string raw_type = node->FirstChildElement("Type")->GetText();
```
trns1997: if `Type` is absent in the ESI file this will be nullptr and crash right?

lib/src/ESI/Parser.cc
```diff
@@ -0,0 +1,667 @@
...
+
+    auto dtype = dtypes_->FirstChildElement();
+    while (dtype)
+    {
+        if (raw_type == dtype->FirstChildElement("Name")->GetText())
```
trns1997: if `Name` is absent in the ESI file this will be nullptr and crash

lib/src/ESI/Parser.cc
```diff
@@ -0,0 +1,667 @@
...
+
+    return dtype;
+}
+
+
+CoE::Object Parser::createObject(XMLNode* node)
+{
+    CoE::Object object;
+    object.index = toNumber<uint16_t>(node->FirstChildElement("Index"));
+    object.name  = node->FirstChildElement("Name")->GetText();
```
trns1997: if `Name` is absent in the ESI file this will be nullptr and crash

trns1997 (approved):

leducp:
Agree with the comment: didn't had the time to check if the fields are mandatory or not

PR #394, opened 2026-05-26
ESI parser step 3: Mailbox + InitCmds

leducp:
Parse <Mailbox> with all six protocols (CoE/EoE/FoE/SoE/AoE/VoE) and their <InitCmd> lists per ESM Transition. 
Protocol blocks exposed as std::optional<…> on the Mailbox aggregate; schema-mandatory <Data>/<Transition> enforced with contextual errors.

### DIFF
--- a/lib/src/ESI/Parser.cc
+++ b/lib/src/ESI/Parser.cc
@@ -154,7 +154,10 @@ namespace
         return buf;
     }
 
-    // xs:boolean per the ESI schema: accepts "true"/"false" and "1"/"0".
+    // xs:boolean per the ESI schema. Returns false when the attribute is
+    // absent (schema default for the ESI attrs we read); throws when present
+    // but not one of the four canonical values, so typos like Virtual="yes"
+    // fail loudly instead of silently being treated as false.
     bool readBoolAttr(XMLElement* node, char const* name)
     {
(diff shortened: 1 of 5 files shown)


lib/src/ESI/Parser.cc
```diff
@@ -168,6 +168,57 @@ namespace
...
+
+    // xs:int attribute: signed decimal, no #x hex prefix. Returns nullopt when absent.
+    std::optional<int32_t> readIntAttr(XMLElement* node, char const* name)
+    {
+        if (node == nullptr)
+        {
+            return std::nullopt;
+        }
+        int32_t value = 0;
+        if (node->QueryIntAttribute(name, &value) != tinyxml2::XML_SUCCESS)
```
trns1997: you nullopt when attribute missing or malformed, but the problem here is that in line https://github.com/leducp/KickCAT/pull/394/changes#diff-44a178f7fa0f0db42609e868d90de266344ea092ea43f12f9c70de5f294890afR523 `cmd.channel = readIntAttr(ic, "Chn").value_or(0);` defaults to 0 even if we have something missing of malformed which does not seem right to me

trns1997 (approved):

PR #395, opened 2026-05-29
Add a generic DS402 drive class

### LINKED ISSUE #396: Add a generic DS402 drive

leducp:
close https://github.com/leducp/KickCAT/issues/396

### DIFF
--- a/examples/master/ds402_drive/CMakeLists.txt
+++ b/examples/master/ds402_drive/CMakeLists.txt
@@ -1,0 +1,7 @@
+if (PIKEOS)
+    set(OS_SRCS ${PROJECT_SOURCE_DIR}/src/OS/PikeOS/p4ext_config.c)
+endif()
+
+add_executable(ds402_sinus ds402_sinus.cc ${OS_SRCS})
+target_link_libraries(ds402_sinus kickcat argparse::argparse)
+set_kickcat_properties(ds402_sinus)
--- a/examples/master/ds402_drive/ds402_sinus.cc
+++ b/examples/master/ds402_drive/ds402_sinus.cc
@@ -1,4 +1,3 @@
-#define _USE_MATH_DEFINES
 #include <cmath>
 #include <iostream>
 
@@ -8,6 +7,7 @@
 #include "kickcat/CoE/CiA/DS402/Drive.h"
 #include "kickcat/Link.h"
 #include "kickcat/OS/Timer.h"
+#include "kickcat/Units.h"
 #include "kickcat/helpers.h"
 
 using namespace kickcat;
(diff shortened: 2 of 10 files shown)


examples/master/ds402_drive/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+if (PIKEOS)
```
trns1997, line 1: do we want to continue pike os support? I am not sure we have the material to validate its working

examples/master/ds402_drive/ds402_sinus.cc
```diff
@@ -0,0 +1,173 @@
...
+    {
+        timer.wait_next_tick();
+        try
+        {
+            bus.sendLogicalRead (bus_error);
+            bus.sendLogicalWrite(bus_error);
+            bus.finalizeDatagrams();
+            bus.processAwaitingFrames();
+        }
+        catch (...) {}
```
trns1997: hmmm this should throw no?

lib/master/src/CoE/CiA/DS402/Drive.cc
```diff
@@ -0,0 +1,160 @@
...
+    {
+        setUnits(units_);
+    }
+
+    Drive::~Drive()
+    {
+        if (out_ != nullptr)
+        {
+            out_->control_word    = control::word::DISABLE_VOLTAGE;
+            out_->target_position = 0;
```
trns1997: why set to 0? I think this should keep last position like we do in attach

examples/master/ds402_drive/ds402_sinus.cc
```diff
@@ -0,0 +1,173 @@
...
+        {
+            bus.sendLogicalRead (bus_error);
+            bus.sendLogicalWrite(bus_error);
+            bus.finalizeDatagrams();
+            bus.processAwaitingFrames();
+        }
+        catch (std::exception const& e)
+        {
+            std::cerr << "[" << i << "] " << e.what() << std::endl;
+            continue;
```
trns1997: shouldn't we `drive.update()` before continue? Do you not want to refresh even if there is a problem?

trns1997 (approved):

examples/master/ds402_drive/CMakeLists.txt
leducp: for now yes: nothing is broken because nothing related was changed

lib/master/src/CoE/CiA/DS402/Drive.cc
leducp: omfg yes bad idea

examples/master/ds402_drive/ds402_sinus.cc
leducp: yes

examples/master/ds402_drive/ds402_sinus.cc
leducp: agree

examples/master/ds402_drive/ds402_sinus.cc
leducp: yes

PR #398, opened 2026-06-02
ESI parser step 4: PDOs + EEPROM + Distributed Clocks

leducp:
Parse <RxPdo>/<TxPdo> with <Entry> children and auto-generate 0x16xx/0x1Axx mapping and 0x1C12/0x1C13 assignment CoE objects into the legacy Dictionary (explicit declarations win). 
Add <Eeprom> (raw or structured with Categories) and <Dc>/<OpMode> (cycle/shift times) as std::optional aggregates on Device.

### DIFF
--- a/lib/src/ESI/Parser.cc
+++ b/lib/src/ESI/Parser.cc
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <cstring>
+#include <limits>
 #include <stdexcept>
 #include <string>
 
@@ -97,8 +98,11 @@ namespace
         return text;
     }
 
-    // Parent + child-name overload: fetches then validates.
-    char const* requireText(XMLNode* parent, char const* child, std::string const& where)
+    // Parent + child-name convenience: fetches the child element, then
+    // delegates to requireText. Distinct name from requireText to avoid an
+    // overload-resolution trap where a derived-class pointer (XMLElement*)
+    // silently binds to the wrong overload.
+    char const* requireChildText(XMLNode* parent, char const* child, std::string const& where)
     {
         return requireText(parent->FirstChildElement(child), child, where);
     }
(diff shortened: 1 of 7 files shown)


lib/src/ESI/Parser.cc
```diff
@@ -565,23 +665,429 @@ void Parser::parseMailbox(XMLElement* device, Mailbox& out)
...
+                char buf[64];
+                std::snprintf(buf, sizeof(buf), "ESI: duplicate %s <Index> 0x%04x", element_name, pdo.index);
+                throw std::invalid_argument(buf);
+            }
+        }
+        out.push_back(std::move(pdo));
+    }
+}
+
+namespace
```
trns1997, line 789: remove anonymous namespacing, i think these can be public apis

unit/kickcat_esi_test_sm_fmmu.xml
```diff
@@ -87,6 +117,32 @@
...
+						<AssignActivate>#x000</AssignActivate>
+					</OpMode>
+				</Dc>
+				<Eeprom AssignToPdi="true">
+					<ByteSize>2048</ByteSize>
+					<ConfigData>05000000F00000000000000000000000</ConfigData>
+					<BootStrap>0010800010108000</BootStrap>
+					<Category>
+						<CatNo>30</CatNo>
+						<DataString>BootCfg</DataString>
```
trns1997, line 143: can this be empty?

trns1997 (approved):

lib/src/ESI/Parser.cc
```diff
@@ -640,22 +1146,49 @@ void Parser::parseFmmus(XMLElement* device, std::vector<Fmmu>& out)
...
     {
-        out.push_back(createObject(node_object));
+        dictionary = profile->FirstChildElement("Dictionary");
+    }
+    if (dictionary != nullptr)
+    {
+        dtypes_       = dictionary->FirstChildElement("DataTypes");
+        // <Objects> is mandatory when <Dictionary> is present per the XSD
+        // (DictionaryType sequence requires it). The whole <Dictionary>
+        // element is optional (loosened earlier in this PR), but if it's
```
trns1997: need to run a pass to remove comments of this nature which bring no value

lib/src/ESI/Parser.cc
leducp: fixed whenever it makes sense

unit/kickcat_esi_test_sm_fmmu.xml
leducp: yes and fixed in the parser

lib/src/ESI/Parser.cc
leducp: done

PR #399, opened 2026-06-03
Hardening

leducp:
Fix perf in PDO hot path
Fix SDO segment access
Better API to catch bugs
Add ASan/LSan/UBsan in CI
Defensive check in gateway

Related to #218


lib/py_bindings/src/bus_py.cc
```diff
@@ -172,7 +172,7 @@ namespace kickcat
             .def("create_mapping", [](Bus &self, int size)
                 {
                     static std::vector<uint8_t> io_buffer(size);
```
trns1997, line 174: is this a good idea? If i call another create mapping with larger size it does not recreate this io_buffer? Not sure how it works with bindings tbh?

lib/src/CoE/mailbox/response.cc
```diff
@@ -136,32 +269,46 @@ namespace kickcat::mailbox::response
...
+            reply(std::move(data_));
+            afterHooks(CoE::Access::READ, entry);
+            return ProcessingResult::FINALIZE;
         }
-        else
+
+        sdo_->transfer_type = 0;
+        std::memcpy(payload_, &size, sizeof(uint32_t)); // complete size
+
+        uint32_t const single_frame_max = static_cast<uint32_t>(data_.size()) - 16;
```
trns1997, line 294: what if data_.size() is < 16?

lib/src/CoE/mailbox/response.cc
```diff
@@ -110,9 +116,136 @@ namespace kickcat::mailbox::response
...
+        if (segmented_entry_ == nullptr)
+        {
+            return ProcessingResult::NOOP; // not serving a segmented upload
+        }
+
+        auto const* header = pointData<mailbox::Header>(raw_message.data());
+        auto const* coe    = pointData<CoE::Header>(header);
+        auto const* sdo    = pointData<CoE::ServiceData>(coe);
+        if ((header->type != mailbox::Type::CoE) or (coe->service != CoE::Service::SDO_REQUEST)
+            or (header->len < 10))
```
trns1997, line 130: why 10? and why no check on raw_message size?

lib/src/CoE/mailbox/response.cc
```diff
@@ -136,32 +269,46 @@ namespace kickcat::mailbox::response
...
+        // to serve the segment requests that follow
+        segmented_entry_  = entry;
+        segmented_offset_ = 0;
+        segmented_toggle_ = false;
         reply(std::move(data_));
-
-        afterHooks(CoE::Access::READ, entry);
-
-        return ProcessingResult::FINALIZE;
+        return ProcessingResult::FINALIZE_AND_KEEP;
```
trns1997, line 311: what is finalize and keep exact? and why change?

lib/src/CoE/mailbox/response.cc
leducp: because what's important is the message len, not the container which is always the mailbox size

lib/src/CoE/mailbox/response.cc
leducp: the message is finalized but more are coming that are associated with the handle so we need to keep it alive

trns1997 (approved):

PR #411, opened 2026-06-15
 KickUI: a GUI tool to inspect, configure and operate an EtherCAT bus

leducp:
KickUI is an imgui/glfw bench tool built on the KickCAT master. It connects to
  a real NIC or to the embedded software simulator (tap), scans the bus and
  shows the discovered topology as a drawn graph (DL_STATUS port states, cable
  break/heal on the simulator), per-slave diagnostics (AL states, error-counter
  totals, event log with alert banner) and a per-device detail view: SDO
  transfers and object-dictionary discovery, PDO mapping read-back and manual
  editing, live process-data bytes, and full DS402 drive control (bring-up,
  state machine, manual/jog/generator setpoints).

### DIFF
--- a/tools/kickui/BusProtocol.h
+++ b/tools/kickui/BusProtocol.h
@@ -1,0 +1,207 @@
+#ifndef KICKCAT_TOOLS_KICKUI_BUS_PROTOCOL_H
+#define KICKCAT_TOOLS_KICKUI_BUS_PROTOCOL_H
+
+#include <atomic>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include "kickcat/CoE/CiA/DS402/Drive.h"   // UnitConfig, ControlMode
+#include "kickcat/CoE/OD.h"                // DataType, ObjectCode, Access
+#include "kickcat/protocol.h"              // State
+
+namespace kickcat::kickui
+{
+    // The data exchanged between the UI thread and the bus-owning actor:
+    // plain value types, the discrete Event stream and the published snapshot.
+    // No live Bus/Link types belong here.
+
+    enum class SetpointSource { Manual, Jog, Generator };
+    enum class Waveform       { Sine, Step, Triangle };
+
+    // Snapshot of the controlled drive, published by the RT thread each cycle
+    // and read by the UI thread.
+    struct DriveFeedback
+    {
+        uint16_t status_word       = 0;
+        uint16_t control_word      = 0;
+        uint16_t error_code        = 0;
+        int8_t   mode_display      = 0;
+        int32_t  actual_pos_raw    = 0;
+        int32_t  actual_vel_raw    = 0;
+        int16_t  actual_torque_raw = 0;
+        double   actual_pos        = 0.0;
+        double   actual_vel        = 0.0;
+        double   actual_torque     = 0.0;
+        double   target            = 0.0;
+        bool     enabled           = false;
+        bool     faulted           = false;
+        bool     operational       = false;
+    };
+
+    // One mapped PDO entry (which object/subindex, and its bit length).
+    struct PdoEntry
+    {
+        uint16_t pdo   = 0;
+        uint16_t index = 0;
+        uint8_t  sub   = 0;
+        uint8_t  bits  = 0;
+    };
+
+    // Per-drive bring-up parameters for operate(). PDO indices default to the
+    // canonical 0x1600/0x1A00 but are overridable (Marvin uses 0x1601/0x1A01).
+    // DC is a bus-wide setting (see setDcConfig), not per-drive.
+    struct OperateConfig
+    {
+        CoE::CiA::DS402::UnitConfig           units;
+        CoE::CiA::DS402::control::ControlMode  mode = CoE::CiA::DS402::control::POSITION_CYCLIC;
+        uint16_t rx_pdo_map  = 0x1600;
+        uint16_t tx_pdo_map  = 0x1A00;
+
+        // Manual PDO mapping: when set, the bring-up writes these entries to the
+        // rx_pdo_map/tx_pdo_map objects (CoE remap sequence) before createMapping
+        // instead of relying on the slave's SII default. Each entry's pdo field is
+        // ignored on apply (the owning object is rx_pdo_map/tx_pdo_map).
+        bool                  manual_mapping = false;
+        std::vector<PdoEntry> manual_rx;   // -> rx_pdo_map / 0x1C12
+        std::vector<PdoEntry> manual_tx;   // -> tx_pdo_map / 0x1C13
+    };
+
+    // Result of one async SDO read/write; the UI polls `done`.
+    struct SdoResult
+    {
+        std::atomic<bool>    done{false};
+        bool                 ok = false;
+        std::string          message;
+        std::vector<uint8_t> data;
+    };
+
+    // One subindex of a RECORD/ARRAY object (its own description + value).
+    struct OdEntry
+    {
+        uint8_t              subindex  = 0;
+        CoE::DataType        data_type = CoE::DataType::UNKNOWN;
+        uint16_t             access    = 0;  // CoE::Access bitmask
+        std::string          name;
+        std::vector<uint8_t> value;
+        std::string          value_error;
+    };
+
+    // One object discovered via the SDO-Information service.
+    struct OdObject
+    {
+        uint16_t             index        = 0;
+        uint8_t              max_subindex = 0;
+        CoE::ObjectCode      object_code  = CoE::ObjectCode::NIL;
+        CoE::DataType        data_type    = CoE::DataType::UNKNOWN;
+        uint16_t             access       = 0;  // CoE::Access bitmask of subindex 0 (from GetED)
+        std::string          name;
+        std::vector<uint8_t> value;        // sub0 value (object value for VAR, count for RECORD/ARRAY)
+        std::string          value_error;  // abort/error text when a value read failed (surfaced to UI)
+        std::vector<OdEntry> entries;      // sub1..N for RECORD/ARRAY (empty for VAR)
+    };
+
+    struct PdoMapping
+    {
+        int                   slave = -1;
+        bool                  valid = false;
+        std::string           error;
+        std::vector<PdoEntry> rx;   // SM2 / 0x1C12 outputs
+        std::vector<PdoEntry> tx;   // SM3 / 0x1C13 inputs
+    };
+
+    // One slave as the master sees it on the wire: its address, the per-port link
+    // state read from DL_STATUS (0x110), and its parent in the discovered tree.
+    struct TopologyNode
+    {
+        int         index          = -1;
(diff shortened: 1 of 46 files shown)


tools/kickui/BusProtocol.h
trns1997, line 1: I think bunch of definitions is this file are already defined in other regions of the existing codebase, I would suggest reusing those defs or maybe adapt them to fit the need?

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2376 @@
...
+#include "kickcat/helpers.h"
+
+using namespace std::chrono;
+using namespace kickcat::CoE::CiA::DS402;
+
+namespace kickcat::kickui
+{
+    namespace
+    {
+        constexpr uint16_t DS402_PROFILE = 402;  // low word of 0x1000 device type
```
trns1997: I think we need to find a way to fill the profile catalogs, that way the bus session is agnostic of the potentially ever growing slave profile list

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2376 @@
...
+                msg += std::string("  [AL status: ") + ALStatus_to_string(al->code()) + "]";
+            }
+            return msg;
+        }
+
+        // Read each slave's DL_STATUS port bits and derive the parent/child tree
+        // (getTopology). Runs on whichever thread currently owns the bus. With
+        // refetch the DL status is re-read from the wire first (live refresh);
+        // at connect time bus.init() has already populated it.
+        TopologyInfo computeTopology(Bus& bus, bool refetch)
```
trns1997, line 44: Do we not already have a method that does this somewhere else? If yes I think we should refactor and mutualise.

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2376 @@
...
+
+    void Device::configureSlave(OperateConfig const& config) { session->configureSlave(index, config); }
+    void Device::includeSlave()                       { session->includeSlave(index); }
+    void Device::includeSlave(OperateConfig const& c) { session->includeSlave(index, c); }
+    void Device::unconfigureSlave()                   { session->unconfigureSlave(index); }
+    bool Device::isConfigured() const                 { return session->isConfigured(index); }
+    bool Device::isOperating() const                  { return session->isOperating(index); }
+    std::shared_ptr<MotorControl> Device::motor() const { return session->motor(index); }
+
+    void BusSession::connect(std::string const& interface)
```
trns1997, line 466: I think the bus latency timeouts should be param, as kickui does not need to be hard realtime, we can have slow masters (non rt kernel) for diagnosis

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2376 @@
...
+
+                    // DS402 = a CoE device whose device-type (0x1000) low word is 402.
+                    if (dev.has_coe)
+                    {
+                        uint32_t device_type = 0;
+                        uint32_t size        = sizeof(device_type);
+                        try
+                        {
+                            bus->readSDO(slave, 0x1000, 0, Bus::Access::PARTIAL, &device_type, &size, 100ms);
+                            dev.is_ds402 = ((device_type & 0xFFFF) == DS402_PROFILE);
```
trns1997: I think we need to rework this, right now we only have ds402 but maybe in the future we might have other profiles

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2376 @@
...
+        if (not connected_ or rt_running_ or operate_requested_)
+        {
+            return;
+        }
+        if ((slave_index < 0) or (slave_index >= static_cast<int>(devices_.size())))
+        {
+            return;
+        }
+        LockGuard lock(controls_mtx_);
+        std::shared_ptr<MotorControl> dc;
```
trns1997: Change DC to motor or something else, I confused it for distributed clock

tools/kickui/BusSession.cc
leducp: But we will still need to own the supported profiles somehow.

tools/kickui/BusSession.cc
leducp: There is and it is used here (cf line 62). This computeToplogy is a wrapper around the tree walk algorithm to add all infos needed by the UI in the answer

tools/kickui/BusSession.cc
leducp: hahaha right now that I re-read it without the context it is quite disturbing

tools/kickui/BusSession.cc
leducp: Well, it is not true: Marvin robot for instance need DC and a proper real time. The UI do need to at least 'try' to reach a proper good enough real time.
About the timeout param though I'll do a check because there is a param.

tools/kickui/BusSession.cc
leducp: Agree: this flag is too specific.

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2412 @@
...
+        // sequence, reusing the lib's mapPDO. No-op for an empty entry list.
+        void writeManualMapping(Bus& bus, Slave& slave, uint16_t pdo_obj,
+                                std::vector<PdoEntry> const& entries, uint16_t sm_map)
+        {
+            if (entries.empty()) { return; }
+            std::vector<uint32_t> words;
+            words.reserve(entries.size());
+            for (auto const& e : entries)
+            {
+                words.push_back((static_cast<uint32_t>(e.index) << 16)
```
trns1997: i think this packing can be refactored and common in protocol.h or OD.h? I feel i've seen us do this in multiple occasions

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2412 @@
...
+                    {
+                        continue;
+                    }
+                    uint32_t entry_count = readU(slave, pdo, 0, 1);
+                    for (uint32_t j = 1; j <= entry_count; ++j)
+                    {
+                        uint32_t e = readU(slave, pdo, static_cast<uint8_t>(j), 4);
+                        PdoEntry pe;
+                        pe.pdo   = pdo;
+                        pe.index = static_cast<uint16_t>(e >> 16);
```
trns1997: lets create a common unpacking method as well here wdyt?

tools/kickui/SdoPanel.cc
```diff
@@ -0,0 +1,536 @@
...
+            {"INTEGER32",       4, true,  false, false, false},
+            {"INTEGER64",       8, true,  false, false, false},
+            {"REAL32",          4, false, false, false, true},
+            {"REAL64",          8, false, false, false, true},
+            {"VISIBLE_STRING",  0, false, true,  false, false},
+            {"RAW (hex)",       0, false, false, true,  false},
+        };
+
+        // Map a CoE DataType (from OD discovery) to a TYPES index; RAW fallback.
+        int typeIndexFor(uint16_t data_type)
```
trns1997: I am not sure why you redefined this and not just reuse what we have in `lib/src/CoE/OD.cc` . Over here you seem to use the library definition already: https://github.com/leducp/KickCAT/pull/411/changes#diff-f4ba7f0fedb97cbf05bb55d564b92ee5b9c0acfa84f1cb034196031b4851269aR500 i feel this can we reworked to remove and reuse CoE existing defs, maybe a tiny helper can be defined if needed. This would avoid any sort of divergence from lib and UI

tools/kickui/BusProtocol.h
```diff
@@ -0,0 +1,206 @@
...
+        bool                 ok = false;
+        std::string          message;
+        std::vector<uint8_t> data;
+    };
+
+    // One subindex of a RECORD/ARRAY object (its own description + value).
+    struct OdEntry
+    {
+        uint8_t              subindex  = 0;
+        uint16_t             data_type = 0;
```
trns1997: we should use library types defined in `lib/include/kickcat/CoE/OD.h`

tools/kickui/BusProtocol.h
```diff
@@ -0,0 +1,206 @@
...
+        uint16_t             access    = 0;
```
trns1997: same here use CoE::Access

tools/kickui/BusProtocol.h
```diff
@@ -0,0 +1,206 @@
...
+        std::string          name;
+        std::vector<uint8_t> value;
+        std::string          value_error;
+    };
+
+    // One object discovered via the SDO-Information service.
+    struct OdObject
```
trns1997, line 91: over here as well?

tools/kickui/Theme.cc
```diff
@@ -0,0 +1,75 @@
...
+    ImVec4 const COLOR_ESM_PREOP{0.36f, 0.60f, 1.00f, 1.0f};   // blue
+    ImVec4 const COLOR_ESM_SAFEOP{0.95f, 0.58f, 0.25f, 1.0f};  // orange
+    ImVec4 const COLOR_ESM_OP{0.26f, 0.82f, 0.48f, 1.0f};      // green
+    ImVec4 const COLOR_ESM_OTHER{0.60f, 0.60f, 0.64f, 1.0f};   // boot/invalid
+
+    char const* stateLabel(uint8_t al_status)
+    {
+        switch (al_status & AL_STATE_MASK)
+        {
+            case static_cast<uint8_t>(State::INIT):        { return "INIT";    }
```
trns1997: we already have a `toString(State)` for this no?

tools/kickui/BusSession.cc
```diff
@@ -0,0 +1,2412 @@
...
+                    link->processDatagrams();
+                }
+                catch (std::exception const&)
+                {
+                    // No answer: treat every slave as a regular ESC.
+                }
+                for (int i = 0; i < static_cast<int>(slaves.size()); ++i)
+                {
+                    auto& slave = slaves[i];
+                    Device dev;
```
trns1997, line 570: I do not get why you create a new device that essentially has the same content as the slave?

trns1997: Ah its a snapshot view, do avoid blockking the bus?

trns1997 (approved):

tools/kickui/BusSession.cc
leducp: Good idea: this is the standard PDO mapping mechanism, it worth its own packing/unpacking in the library!

tools/kickui/SdoPanel.cc
leducp: right: I slowly build another method to do so while iterating... let's refactor!

tools/kickui/Theme.cc
leducp: not for short string. However should be promoted in the lib!

tools/kickui/BusSession.cc
leducp: This is the UI counterpart of the RT loop slave.

Issue #397, opened 2026-06-01
Device with no Dictionary and no full Assignment Object


phamnhatha090805:
Hi sir,
I'm trying to use KickCAT to simulate the Beckhoff EtherCAT modules. In this case it's the EL1004.
Unlike some other devices, this ESI files provide PDO information through TxPdo sections but do not contain a complete CoE dictionary. Because of this the flag `is_mapped` will never be set to TRUE.
I don't know if you gus are already awared of this.

leducp:
I'm not aware of this bug no. Can you send me the EL1004 ESI and the way you are currently using it so I can have a look?

phamnhatha090805:
> I'm not aware of this bug no. Can you send me the EL1004 ESI and the way you are currently using it so I can have a look?

Sure. Attached is the EL1xxx ESI XML file. The EL1004 section is representative of the issue I'm seeing.
The EL1004 only defines Fixed TxPDO mappings and does not provide a complete CoE object dictionary with the corresponding PDO entries. In my use case with KickCAT, I need to access the PDO data of each virtual slave through the CoE dictionary. However, because the ESI only contains the fixed PDO mapping information, KickCAT is currently unable to create the corresponding PDO entries in its object dictionary, which prevents PDO access and mapping.

[Beckhoff EL1xxx.xml](https://github.com/user-attachments/files/28502293/Beckhoff.EL1xxx.xml)

phamnhatha090805:
Hi,
I'm testing the recent ESI parser changes from PR #398 and trying to simulate a Beckhoff EL1004 terminal using the network simulator 
Setup:  I run the KickCAT on the Rasp Pi and I pass 2 EtherCAT device to the network_simulator EK1100 (EtherCAT bus coupler) -> EL1004 

The problem is that the entry remains: `<is_mapped = false>` for the entire runtime even the TwinCAT showing all devices move to OP state.
Is there currently any limitation in the new ESI-generated dictionary path for terminals such as EL1004 that only define PDOs in the ESI?
Should the generated PDO target objects (0x6000, 0x6010, ...) be marked by PDO::configureMapping() automatically, or is there an additional step required to make these synthesized entries participate in PDO mapping?
Any guidance would be appreciated.

leducp:
Hi
It is still a work in progress. The last PR is not merged yet (@trns1997 didn't had the time to review it) but this is the one that should parse and populate the dictionary from the PDO section if no dictionary provided by the ESI. Did you try with this one?

[...]

PR #425, opened 2026-06-30
Use SII mapping when there is no mailbox (simple slaves)

leducp:
related to #397

### DIFF
--- a/tools/kickui/BusSession.cc
+++ b/tools/kickui/BusSession.cc
@@ -25,6 +25,34 @@ namespace kickcat::kickui
 {
     namespace
     {
+        // A mailboxless terminal's PDO assignment is fixed and fully described by
+        // the SII PDO categories, so build the read-back view straight from them
+        // (no SDO, nothing to poll on the wire).
+        PdoMapping siiPdoMapping(int slave_index, eeprom::SII const& sii)
+        {
+            auto fromSII = [](std::vector<eeprom::PDOMapping> const& pdos, std::vector<PdoEntry>& out)
+            {
+                for (auto const& pdo : pdos)
+                {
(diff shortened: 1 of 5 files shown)


trns1997 (approved):

tools/kickui/BusSession.cc
```diff
@@ -25,6 +25,34 @@ namespace kickcat::kickui
 {
     namespace
     {
+        // A mailboxless terminal's PDO assignment is fixed and fully described by
+        // the SII PDO categories, so build the read-back view straight from them
+        // (no SDO, nothing to poll on the wire).
+        PdoMapping siiPdoMapping(int slave_index, eeprom::SII const& sii)
+        {
+            auto fromSII = [](std::vector<eeprom::PDOMapping> const& pdos, std::vector<PdoEntry>& out)
```
windelbouwman, line 33: This is an interesting function. Is this something that also could be used in the `SimulatedSlave` as well? Then the `is_mapped` flag can also be properly set there.

leducp: Probably yes :) 
To be noted though that is_mapped is dedicated to handle aliasing in the dictionary, so the slave stack can handle it properly. The "problem" here is that EL1004 is a simple slave, probably a pure electronic one (because emulated). The dictionary/slave stack as no signification for it: you can either use it and hack so it works (working around the FSM emulation, the dictionary aliasing etc) or you can just work on top of the ESC directly like a simple slave do (read/write the ESC mapping through the SII).

I think it would be better to leverage a more stupid simple (KISS) SimulatedSlave that just does the PDO job. The nuttx arduino-due slave example may be a good fit to see how to do it: no mailbox, no dictionary, just the SII and a really simple main file.

PR #426, opened 2026-07-01
Use product code in eeprom to find device in ESI file in SimulatedSlave.

windelbouwman:
I added a verification between eeprom and coe_xml file, such that the product code and revision number are read from the eeprom, and then the device is searched in the coe_xml file.

### DIFF
--- a/examples/slave/nuttx/lan9252/freedom-k64f/freedom-k64f.xml
+++ b/examples/slave/nuttx/lan9252/freedom-k64f/freedom-k64f.xml
@@ -16,7 +16,7 @@
     </Groups>
     <Devices>
       <Device Physics="YY">
-        <Type ProductCode="#x00000001" RevisionNo="#x00000001">Board</Type>
+        <Type ProductCode="#x00DEFEDE" RevisionNo="#x00005A01">Board</Type>
         <Name LcId="1033">Board</Name>
         <Info>
           <StateMachine>
--- a/lib/simulation/src/SimulatedSlave.cc
+++ b/lib/simulation/src/SimulatedSlave.cc
@@ -9,21 +9,43 @@
 #include <nlohmann/json.hpp>
 
 #include "kickcat/CoE/OD.h"
+#include "kickcat/ESC/EmulatedESC.h"
+#include "kickcat/ESI/Parser.h"
 #include "kickcat/ESI/SIIBuilder.h"
+#include "kickcat/SIIParser.h"
 
 namespace kickcat::sim
 {
     using json = nlohmann::json;
     namespace fs = std::filesystem;
 
-    SimulatedSlave buildSlave(std::string const& config_path, ESI::Parser& parser)
+    void configureDeviceDictionary(SimulatedSlave& sim, ESI::Device& device)
     {
-        fs::path config_dir = fs::path(config_path).parent_path();
+        if (not device.dictionary.empty())
+        {
+            sim.dictionary = std::make_unique<CoE::Dictionary>(std::move(device.dictionary));
+            sim.slave->setDictionary(sim.dictionary.get());
+
+            // A mailboxless terminal (e.g. a digital I/O like EL1004) still gets its OD,
+            // but only a device that declares a CoE mailbox is reachable by SDO.
+            const bool esi_coe_advertised = (device.mailbox and device.mailbox->coe);
+            if (esi_coe_advertised)
+            {
+                sim.mailbox = std::make_unique<mailbox::response::Mailbox>(sim.esc.get(), 1024);
+                sim.mailbox->enableCoE(*sim.dictionary);
+                sim.slave->setMailbox(sim.mailbox.get());
+            }
+        }
+    }
+
+    SimulatedSlave buildSlave(fs::path const& config_path)
+    {
+        fs::path config_dir = config_path.parent_path();
 
         std::ifstream f(config_path);
         if (not f.is_open())
         {
-            throw std::runtime_error("Failed to open config file: " + config_path);
+            throw std::runtime_error("Failed to open config file: " + config_path.string());
         }
         json config;
         try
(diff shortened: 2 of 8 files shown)


simulation/slave_configs/xmc4800.json
```diff
@@ -1,4 +1,4 @@
 {
-  "eeprom": "../../examples/slave/nuttx/xmc4800/boards/relax/eeprom_xmc4800.bin",
+  "eeprom": "../../examples/slave/nuttx/xmc4800/boards/wdc_foot/foot.bin",
```
windelbouwman: I had to change this bin file, otherwise this eeprom image does not match xmc4800.xml.

leducp:
Sorry for the delay didn't have the time to test/try/play/read with this PR. I hope to be able to do it this weekend!

examples/slave/nuttx/lan9252/freedom-k64f/freedom-k64f.xml
```diff
@@ -16,7 +16,7 @@
     </Groups>
     <Devices>
       <Device Physics="YY">
-        <Type ProductCode="#x00000001" RevisionNo="#x00000001">Board</Type>
```
leducp, line 19: unrelated change?

CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ if (NUTTX)
   set(BUILD_EEPROM_EDITOR   OFF CACHE BOOL "" FORCE)
 endif()
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
leducp: unrelated change

tools/kickui/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_executable(kickui
 target_link_libraries(kickui PRIVATE
     kickcat
     kickcat_imgui
+    portable-file-dialogs
```
leducp, line 20: I think this patch was merged in master (may conflict)

simulation/slave_configs/xmc4800.json
leducp: I think it is worth to:
- duplicate xmc4800.xml (ESI) and change its reference code to match the one in the foot
- edit xmc4800.xml to match the product code of the xmc480.bin
- move both esi in their respective board

With this all boards are coherents and can be simulated

lib/simulation/src/SimulatedSlave.cc
```diff
@@ -9,21 +9,42 @@
...
 #include "kickcat/ESI/SIIBuilder.h"
+#include "kickcat/SIIParser.h"
 
 namespace kickcat::sim
 {
     using json = nlohmann::json;
     namespace fs = std::filesystem;
 
-    SimulatedSlave buildSlave(std::string const& config_path, ESI::Parser& parser)
+    void configureDeviceDictionary(SimulatedSlave& sim, ESI::Device& device) {
```
leducp: allman (line feed before { )

lib/simulation/src/SimulatedSlave.cc
```diff
@@ -80,41 +94,33 @@ namespace kickcat::sim
...
+                    throw std::runtime_error("CoE XML file not found: " + coe_xml_full_path.string());
+                }
+                eeprom::SII sii;
+                sii.parse(eeprom_image);
+                uint32_t revision_no = sii.info.revision_number;
+                uint32_t product_code = sii.info.product_code;
+
+                ESI::DeviceFilter filter;
+                filter.revision_no = revision_no;
+                filter.product_code = product_code;
```
leducp, line 115: Since we are more strict now (and more compliant I would say) I think we need to evolve a bit the filter to accept esi.revision <= **eeprom**.revision

Could be done in a following PR or recorded in a ticket. Wdyt?

leducp:
Good overall: a bit of refine and we can merge :)

simulation/slave_configs/xmc4800.json
windelbouwman: Okay, that sounds logical. Both relax and wdc_foot should have a bin file, and an esi file. Another option is to have a single ESI file describing both boards, since the parser now supports this, and with the filter we can extract the right dictionary.

In any case, could you clarify what the vendor/product ID should be of the ESI files? I'm not aware of these examples, what they are.

lib/simulation/src/SimulatedSlave.cc
windelbouwman: I do not fully understand, should revision not exactly match? Is it a match if the revision number in the ESI is higher than the revision number in the EEPROM?

Anyway I would change this in a separate PR.

examples/slave/nuttx/lan9252/freedom-k64f/freedom-k64f.xml
windelbouwman: No, I had to change it for the test cases to pass. It now matches with eeprom.bin file. I checked that file with the EEPROM Editor.

lib/simulation/src/SimulatedSlave.cc
leducp: Mmm I'll check with the doc: my understanding is that it is the 'patch' version so if the eeprom revision is greater the ESI revision may be accepted (and my comment have a nice typo... fixed)

simulation/slave_configs/xmc4800.json
leducp: About the vendor/product ID, the eeprom stored in the git is the source of truth and the ESI should be aligned on it (as you already did on the k64f) :)
I would prefer to split the ESI: wdc_foot is a device from a company that run on a PCB with an xmc4800, but it is not a xmc4800-relax board per see (even if you can tweak it so it works on it)

lib/simulation/src/SimulatedSlave.cc
leducp: Anyway, I'll open a ticket to track it : not related to this PR

simulation/slave_configs/xmc4800.json
windelbouwman: The foot already has foot.bin and corresponding foot.xml. The id's there match, so that is ok. I moved the xmc4800.xml into the relax folder, and adjusted the product id and revision number to match the eeprom.bin in the relax folder.

trns1997 (approved):

leducp (approved):

PR #429, opened 2026-07-16
Add a PLL mechanism to lock the master on the DC ref clock

leducp:
* OS/Time: make seconds_f a double to avoid long-run precision loss
* marvin: 1 Hz PLL/DC telemetry with smoothed drift

### DIFF
--- a/lib/src/OS/SoftPll.cc
+++ b/lib/src/OS/SoftPll.cc
@@ -1,0 +1,159 @@
+#include "kickcat/OS/SoftPll.h"
+#include "kickcat/OS/math.h"
+
+namespace kickcat
+{
+    SoftPll::SoftPll(nanoseconds cycle_period)
+        : SoftPll(cycle_period, Config{})
+    {
+    }
+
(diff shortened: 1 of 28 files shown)


lib/src/OS/SoftPll.cc
```diff
@@ -0,0 +1,159 @@
...
+        {
+            error += cycle_;
+        }
+        last_error_ = error;
+        ++samples_;
+
+        // Coast (don't learn) on an overrun or a post-lock outlier. The outlier gate is armed only
+        // after first lock: pre-lock a large error is the startup offset to slew in, not noise.
+        bool reject = cycle_overran;
+        if (ever_locked_ and abs_value(error) > reject_)
```
trns1997, line 76: `reject_` never update but `reject` is used i am a bit confused on the existence of this variable which one is for what is not clear?

leducp: reject_ is a threshold, reject is a local variable to record that the sample was rejected or not to act on it

trns1997 (approved):

PR #435, opened 2026-09-23
FoE: Add core support

leducp:
* Master mailbox
* Slave mailbox
* Simulator


trns1997:
What was this tested on?

lib/include/kickcat/FoE/Storage.h
trns1997, line 1: I do not understand why you need these abstract interface classes? I mean I have a hard time understanding what other storage interfaces we can have. Also foe reader is Specific to foe it would make sense to have abstract reader higher up and then have for reader derive from that unless I am missing something, because all we do is read files over éthercat not sure what else is there

leducp: WIth FoE, you can read and write from/to a slave and from/to the master. Theses classes are here to abstract how you can do it on a uC or on a full fledge OS. 
A file abstraction can do the job on a standard platform (macOS, Linux, Windows) but not on a uC where a file does not always makes sense (i.e. raw flash access in the uC, eeprom through I2C/SPI, etc)

trns1997 (approved):

